PR #10, opened 2024-07-10
Android refactor

graszka22:
What was done:
- introduced a command queue, similar to ts-client, to avoid adding new tracks while renegotiation is in progress etc.
- cleaned up tracks and tracks id - now there is just one track id that always works and the native side is the source of truth what endpoints and tracks we have, rn client and users can always query android-client with `getRemotePeers` and  `getLocalEndpoint` just like in ts-client
- removed remains of membrane rtc 
- simplified rn client android logic
- fixed issue with displaying preview and video view of local track at once - now there is just one local track for both preview and sent local track and it's started by `startCamera`
- simplified screenshare - now user manages foreground service, for example using notifee library, just like foreground service for camera. Previously there might have been two notifications - one for screenshare and one for camera


packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import kotlinx.coroutines.Job
+
+internal class CommandsQueue {
+  private var commandsQueue: ArrayDeque<Command> = ArrayDeque()
+  var clientState: ClientState = ClientState.CREATED
+
+  fun addCommand(command: Command): Job {
+    commandsQueue.add(command)
+    if (commandsQueue.size == 1) {
```
karkakol, line 11: why we only execute command if queueSize is equal to one?

examples/video-chat/screens/PreviewScreen/index.tsx
```diff
@@ -65,27 +69,41 @@ const PreviewScreen = ({ navigation, route }: Props) => {
...
 
   useEffect(() => {
     getCaptureDevices().then((devices) => {
       availableCameras.current = devices;
+
+      startCamera({
+        simulcastConfig: {
+          enabled: true,
+          activeEncodings:
+            Platform.OS === 'android' ? ['l', 'm', 'h'] : ['l', 'h'],
```
mironiasty, line 91: Does anyone knows why there is different set for ios and android?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  fun finishCommand() {
+    val command = commandsQueue.first()
+    val job = command.job
+    commandsQueue.removeFirst()
+    if (command.clientStateAfterCommand != null) {
+      clientState = command.clientStateAfterCommand
+    }
+    job.complete()
+    // TODO: make it iterative, not recursive?
```
mironiasty, line 25: Could we have any task for this TODO with some explanation?

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoRendererView.kt
```diff
@@ -1,26 +1,24 @@
...
   appContext: AppContext
 ) : ExpoView(context, appContext),
   RNFishjamClient.OnTrackUpdateListener {
-  var isInitialized = false
   var activeVideoTrack: VideoTrack? = null
   var trackId: String? = null
 
   private val videoView =
-    VideoTextureViewRenderer(context).also {
+    RNFishjamClient.fishjamClient!!.createVideoViewRenderer().also {
```
mironiasty, line 21: Should we have some check here to ensure `fishjamClient` is not null? Or maybe use different types to avoid such code

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
```diff
@@ -498,7 +517,7 @@ internal class PeerConnectionManager(
           return@launch
         }
       }
-      peerConnectionListener.onAddTrack(trackId!!, receiver!!.track()!!)
+      listeners.forEach { listener -> listener.onAddTrack(trackId!!, receiver!!.track()!!) }
```
mironiasty, line 520: Could we refactor code to avoid `!!`?

examples/video-chat/screens/PreviewScreen/index.tsx
graszka22: yes, I know why
will add a comment in the code or the docs
on iOS on some devices if using H264 hardware encoder there is a hardware limit of 3 encoders (https://github.com/twilio/twilio-video-ios/issues/17) 
every simulcast layer is an encoder
if you start screencast that's one more encoder
3 simulcast layers + 1 screencast layer = 4 encoders
4 > 3, so screencast is going to silently break

afaik android has no such limitations

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClient.kt
```diff
@@ -117,7 +120,7 @@ class FishjamClient(
    * @param trackId an id of a valid local track that has been created using the current client
    * @return a boolean whether the track has been successfully removed or not
    */
-  fun removeTrack(trackId: String): Boolean = client.webrtcClient.removeTrack(trackId)
+  suspend fun removeTrack(trackId: String) = client.removeTrack(trackId)
```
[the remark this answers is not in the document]
graszka22: Because imo it's not useful to return a boolean here and I don't even know why it was here in the first place. If user gets returned false here, what are they supposed to do? Try again or what?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -1,175 +1,729 @@
...
-            response: Response
-          ) {
-            listener.onSocketOpen()
-            val authRequest =
-              PeerMessage
-                .newBuilder()
-                .setAuthRequest(PeerMessage.AuthRequest.newBuilder().setToken(config.token))
-                .build()
-            sendEvent(authRequest)
+      // temporary workaround, the backend doesn't add ~ in sdp answer
```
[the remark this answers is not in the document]
graszka22: it's probably still valid

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionFactoryWrapper.kt
```diff
@@ -0,0 +1,125 @@
...
+      rootEglBase = eglBase,
+      videoParameters = videoParameters,
+      captureDeviceName
+    )
+
+  fun createAudioSource(): AudioSource {
+    if (ContextCompat.checkSelfPermission(appContext, Manifest.permission.RECORD_AUDIO) !=
+      PackageManager.PERMISSION_GRANTED
+    ) {
+      throw SecurityException("Missing permissions to start recording the audio")
```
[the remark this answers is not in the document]
graszka22: I think it shouldn't be handled. If user tries to create an audio track without record audio permission, we should fail hard and user should add the permission handling.

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoRendererView.kt
graszka22: I added `lateinit` to `fishjamClient`. We're sure it's initialized, because it's initialized when the react-native-client module is created. And in order for the user to call anything, the module must be created. So `lateinit` sounds good here.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
graszka22: No, we probably can't. If something is null here that probably means something went very wrong and the backend gave us wrong data, so it's going to crash anyway later and we can't reasonably handle it. Let's crash earlier, so it's easier to debug.

graszka22:
- [ ] fix android 14 screencast
- [ ] finish jobs with an exception on queue clear

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/Command.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.coroutines.CompletableJob
+import kotlinx.coroutines.Job
+
+internal enum class CommandName {
+  CONNECT,
+  JOIN,
+  ADD_TRACK,
+  REMOVE_TRACK,
+  RENEGOTIATE,
+  LEAVE
```
karkakol, line 12: Are those used anywhere?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -1,175 +1,729 @@
...
+      for ((trackId, trackData) in it.tracks) {
+        val track = Track(null, it.id, trackId, trackData.metadata ?: mapOf())
+        endpoint = endpoint.addOrReplaceTrack(track)
+
+        this.listener.onTrackAdded(track)
+      }
+      this.remoteEndpoints[it.id] = (endpoint)
+    }
+    listener.onJoined(endpointID, remoteEndpoints)
+    commandsQueue.finishCommand()
```
karkakol, line 189: What ensures that first command is Connected?
If i am not mistaken sb can call join on empty queue, then first command is `Join`, not `Connected`

yarn.lock
```diff
@@ -3209,12 +3209,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@notifee/react-native@npm:^7.8.2":
+"@notifee/react-native@https://gitpkg.vercel.app/invertase/notifee/packages/react-native?9957a6a8e4f5b9d95c3418705e2ed7a5ad6dfb19&scripts.postinstall=genversion%20--es6%20--semi%20src%2Fversion.ts%20%26%26%20tsc%20%7C%7C%20exit%200":
```
mironiasty, line 3212: Why are we using this version of notifee? (instead of official release)

examples/video-chat/screens/PreviewScreen/index.tsx
mironiasty: In that case, I think we should make this change on SDK level 🤔

scripts/init.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 brew install swift-format ktlint release-it
+npm install -g genversion typescript
```
mironiasty, line 3: Why we need global install of `genversion` and `typescript`?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
mironiasty: Hmm, I think we should not crash app in case of incorrect data. We can throw some error back to app, but crashing is something we should really avoid. (crashing app is worse UX than just showing error)

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
graszka22: That somebody is us, in FishjamClientInternal. User just calls `connect`, we are calling connect and then join.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
graszka22: https://linear.app/swmansion/issue/FCE-303/make-commands-queue-iterative-instead-of-recursive

this is very low priority though

mironiasty (approved):

karkakol (approved):
👍

PR #11, opened 2024-07-10
[FCE-259] Remove ios example

karkakol:
## Description

Remove iOS example and files related to it.


packages/ios-client/project.json
```diff
@@ -0,0 +1,222 @@
+{
```
graszka22, line 1: is this file needed?

karkakol: This package is created by script `gen-mocks.sh`. It is necessary for iOS tests. Not sure though is necessary in repo. On the other hand it affects testing soo i think it should be in repo. If you think the opposite give me your thoughts.

karkakol: Never mind. Deleted.

graszka22 (approved):

PR #12, opened 2024-07-11
[FCE-268] Rename methods

mironiasty:
## Description

Small refactor to remove methods overwrites in `index.ts` file. Also move types to better files (for components)

## Motivation and Context

This is small step to clean up TypeScript code in our SDK

## How has this been tested?

This is safe refactor, it is enough to make sure TS check is still passing

## Types of changes

      not work as expected)

## Checklist:


karkakol (approved):
👍

packages/react-native-client/src/common/metadata.ts
```diff
@@ -1,14 +1,13 @@
...
-/**
 /**
  * a function that updates endpoints's metadata on the server
  * @param metadata a map `string -> any` containing user's track metadata to be sent to the server
  */
-export async function updateEndpointMetadata<
-  EndpointMetadataType extends Metadata,
->(metadata: EndpointMetadataType) {
+export async function updatePeerMetadata<EndpointMetadataType extends Metadata>(
+  metadata: EndpointMetadataType,
```
graszka22, line 9: `EndpointMetadataType`?

packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -8,42 +8,44 @@ import {
...
+
 /**
- * This hook provides live updates of room endpoints.
- * @returns An array of room endpoints.
+ * This hook provides live updates of room peers.
+ * @returns An array of room peers.
  */
-export function useEndpoints<
-  EndpointMetadataType extends Metadata,
+export function usePeers<
```
graszka22, line 21: Notice that they're going to be components like here: https://github.com/fishjam-cloud/web-client-sdk/blob/main/packages/ts-client/src/FishjamClient.ts#L537 (they're going to be soon in mobile?)
I still don't quite understand the difference between components and peers and how user handles it, but probably there is going to be a lot of code reuse.
In particular if user handles componenets and peers in the same way then maybe there is no point in making a distinction between peers and components and we should just leave the endpoints?

PR #14, opened 2024-07-11
Update action names

mironiasty:
## Description

Update action names, so it will be easier to set up repository configuration

## Motivation and Context

Just make life easier


.github/workflows/react_native_lint.yaml
karkakol, line 1: change name to all_packages_lint.yaml

karkakol (approved):
👍

PR #18, opened 2024-07-12
[FCE-273] Decompose global files with types

mironiasty:
## Description

Move almost all types from global `RNFishjamModule.types.ts` into files that these types are actually used. Also add explicit exports in index.ts file. 

Rename `RNFishjameClientModule.ts` into `FishjamClient.ts` (comments on this change are welcome, I'm not sure about it)

Rename `RNFishjamModule.types.ts` into `types.ts` -> new file has only few global types now, that are used across various files.

## Motivation and Context

We had one global file with types. And this is not optimal, as there is no way to verify if these global types are still used and where are they used. By keeping most of types, close to their usage, it is easier to track how they are used.

Also right now we have better control on what types are exported from library, so we know which parts are public and can be used by 3rd party SDK users.

## How has this been tested?

This is only TS change, so only static checks should be verified.

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -1,13 +1,90 @@
...
+ * - ENCODING_INACTIVE - previously selected encoding became inactive
+ * - LOW_BANDWIDTH - there is no longer enough bandwidth to maintain previously selected encoding
+ */
+export enum EncodingReason {
+  Other = 'other',
+  EncodingInactive = 'encodingInactive',
+  LowBandwidth = 'lowBandwidth',
+}
+
+export type EndpointsUpdateEvent<
```
karkakol, line 49: Rename to `PeersUpdateEvent` ?

packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -1,13 +1,90 @@
...
+  AudioTrackMetadataType extends Metadata,
+> = {
+  EndpointsUpdate: Endpoint<
+    MetadataType,
+    VideoTrackMetadataType,
+    AudioTrackMetadataType
+  >[];
+};
+
+export type Endpoint<
```
karkakol, line 61: Rename to `Peer` or even remove?

packages/react-native-client/src/stats/useRTCStatistics.ts
```diff
@@ -1,12 +1,8 @@
...
 import { useCallback, useEffect, useState } from 'react';
 
-import {
-  RTCInboundStats,
-  RTCOutboundStats,
-  RTCStats,
-} from '../RNFishjamClient.types';
-import RNFishjamClientModule from '../RNFishjamClientModule';
+import type { RTCInboundStats, RTCOutboundStats, RTCStats } from './types';
+import RNFishjamClientModule from '../FishjamClient';
```
karkakol, line 5: I do not like the way how we once use `RNFishjamClientModule` and `FishjamClient`. I think it should be unified.

packages/react-native-client/src/hooks/usePeers.ts
mironiasty: I wanted to make this rename in separate PR. But maybe it make sense to do it now

packages/react-native-client/src/hooks/usePeers.ts
mironiasty: same as above

packages/react-native-client/src/stats/useRTCStatistics.ts
mironiasty: I will revert this change, and maybe do it in separate PR (as it would indeed require more changes

packages/react-native-client/src/hooks/usePeers.ts
mironiasty: I was thinking about it - but to update Endpoint to Peers everywhere, i would require more changes (also in native code, due to events). There is another ticket to do this ([FCE-212](https://linear.app/swmansion/issue/FCE-212/removerename-references-to-endpoint-from-rn-sdk-api)), so let's leave old names for now - and update them with new task.

karkakol: 👍

packages/react-native-client/src/hooks/usePeers.ts
karkakol: 👍

packages/react-native-client/src/stats/useRTCStatistics.ts
karkakol: 👍 
Create ticket for it -> approve

karkakol (approved):
👍

PR #21, opened 2024-07-15
[FCE-274] Remove enums from SDK

mironiasty:
## Description

Remove `enum`s from TypeScript code. And use `union`s instead. 

Fix bug, where `EncodingReason` had incorrect mapping to native part.

## Motivation and Context

Enums bad. Unions good.

### DIFF
--- a/examples/video-chat/hooks/useJoinRoom.ts
+++ b/examples/video-chat/hooks/useJoinRoom.ts
@@ -1,8 +1,4 @@
-import {
-  useCamera,
-  useMicrophone,
-  VideoQuality,
-} from '@fishjam-cloud/react-native-client';
+import { useCamera, useMicrophone } from '@fishjam-cloud/react-native-client';
 import notifee, {
   AndroidImportance,
   AndroidColor,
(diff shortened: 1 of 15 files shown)


examples/video-chat/hooks/useJoinRoom.ts
karkakol, line 1: fix warning

examples/video-chat/screens/RoomScreen.tsx
karkakol, line 1: fix warning

karkakol (commented):
It would be nice to unify types string casing. Once we use `export type VideoLayout = 'FILL' | 'FIT'`, in other place `export type VadStatus = 'silence' | 'speech';`. But I am not sure there exists eslint rule for it

mironiasty:
> It would be nice to unify types string casing. Once we use export type VideoLayout = 'FILL' | 'FIT', in other place export type VadStatus = 'silence' | 'speech';. But I am not sure there exists eslint rule for it

Maybe, but I would prefer to do it in separate PR. This change would require native code changes + and much more testing

examples/video-chat/hooks/useJoinRoom.ts
mironiasty: I've removed warnings + I've updated eslint config to fail in case of warnings in future.

karkakol (approved):

PR #33, opened 2024-07-23
[FCE-311] [FCE-320] android front camera mirror

karkakol:
Fix front camera mirroring and toggle function for android


packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
```diff
@@ -101,20 +97,28 @@ class CameraCapturer(
   }
 
   fun flipCamera() {
+    val enumerator = getEnumerator(context)
+    val wasFrontFacing = if (cameraName != null) enumerator.isFrontFacing(cameraName) else false
+
+    for (name in enumerator.deviceNames) {
+      if (wasFrontFacing && enumerator.isBackFacing(name) || !wasFrontFacing && enumerator.isFrontFacing(name)) {
```
graszka22, line 104: please at least add some () for readability

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
```diff
@@ -101,20 +97,28 @@ class CameraCapturer(
...
+      }
+    }
+    setMirrorVideo(!wasFrontFacing)
     cameraCapturer.switchCamera(this)
   }
 
   fun switchCamera(deviceName: String) {
+    val enumerator = getEnumerator(context)
+    cameraName = deviceName
+    setMirrorVideo(enumerator.isFrontFacing(deviceName))
```
graszka22, line 116: are you sure that the front camera is always mirrored? 
I know it's very common and normal but what if user decides otherwise?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
```diff
@@ -18,7 +17,7 @@ class LocalVideoTrack(
   mediaTrack: org.webrtc.VideoTrack,
   endpointId: String,
   metadata: Metadata,
-  private val capturer: Capturer,
+  public val capturer: Capturer,
```
graszka22, line 20: are you sure you want this to be `public` so that user can see it and mess with it? 
maybe at least `internal`?

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoRendererView.kt
```diff
@@ -16,19 +26,39 @@ class VideoRendererView(
...
     activeVideoTrack = videoTrack
+
+    if (videoTrack is LocalVideoTrack && videoTrack.capturer is CameraCapturer) {
+      (videoTrack.capturer as CameraCapturer).setMirrorVideo = { isFrontCamera ->
+        setMirrorVideo(isFrontCamera)
+      }
+      setMirrorVideo(getEnumerator(context).isFrontFacing((videoTrack.capturer as CameraCapturer).cameraName))
+    }
+
+    videoTrack.addRenderer(videoView)
```
graszka22, line 61: this is the same code as in Preview, can we reuse it somehow? or add it to the VideoViewRenderer from android client?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
karkakol: If it would be internal i would not see it in react-native android code

graszka22: the general idea is that the rn-client is like android-client user
so if you need something private from android-client to implement something in rn then it's better to implement something in android-client (if possible)

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoRendererView.kt
karkakol: `VideoTextureViewRenderer` does not have capturer as parameter. So it would not be possible to determine camera properties.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
graszka22: ok, correct me if I'm wrong but what if:
- user sets camera to be not mirrored
- switches from back camera to front camera
- setMirrorVideo is called with `true` because it's now a front camera and it will be mirrored but mirror property is still false?

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoRendererView.kt
graszka22: but you have video track
what if video track had a method `setMirrorVideo`? and the video track would interact with its private capturer?

karkakol: VideoTextureViewRendered does not have videotrack, correct me if i am mistaken

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/LocalVideoTrack.kt
karkakol: now there are two seperate variables, one for mirror and one for if front camera

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/VideoPreviewView.kt
```diff
@@ -1,25 +1,32 @@
...
 import expo.modules.kotlin.AppContext
-import expo.modules.kotlin.views.ExpoView
+import kotlinx.coroutines.cancel
 import org.webrtc.RendererCommon
 
 class VideoPreviewView(
   context: Context,
   appContext: AppContext
-) : ExpoView(context, appContext) {
+) : MirrorableView(context, appContext) {
```
graszka22, line 18: Using inheritance for this case doesn't feel nice but I don't know how to explain it 😅 in kotlin you can inherit after a single class, what if we want to inherit after something else?  Maybe VideoPreviewView should have a reference to something like MirrorViewAnimator?

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/MirrorableView.kt
```diff
@@ -0,0 +1,51 @@
...
+  ) {
+    if (this.mirrorVideo == mirrorVideo && this.isFrontCamera == isFrontCamera) return
+    this.mirrorVideo = mirrorVideo ?: this.mirrorVideo
+    this.isFrontCamera = isFrontCamera ?: this.isFrontCamera
+    if (!isInitialized) return
+
+    coroutineScope.launch {
+      fadeAnimation.start()
+      delay(200)
+      videoView?.setMirror(this@MirrorableView.mirrorVideo xor this@MirrorableView.isFrontCamera)
```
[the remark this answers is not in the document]
karkakol: It is because:
mirror | isFront | shouldReallyMirror:
T.        | T           | F
T.        | F           | T
F.        | T           | T
F.        | F           | F
and this is exactly xor 
if you want i can add comment with documentation

examples/video-chat/screens/PreviewScreen/index.tsx
```diff
@@ -131,12 +133,20 @@ const PreviewScreen = ({ navigation, route }: Props) => {
...
+          />
         ) : (
           <NoCameraView username={route?.params?.userName || 'RN Mobile'} />
         )}
       </View>
       <View style={styles.mediaButtonsWrapper}>
+        <InCallButton
+          iconName={mirrorPreview ? 'keyboard-tab' : 'keyboard-tab-reverse'}
+          onPress={() => setMirrorPreview((prev) => !prev)}
+          accessibilityLabel={TOGGLE_MICROPHONE_BUTTON}
```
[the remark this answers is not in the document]
karkakol: Skill issue.

examples/video-chat/screens/PreviewScreen/index.tsx
```diff
@@ -131,12 +133,20 @@ const PreviewScreen = ({ navigation, route }: Props) => {
     <SafeAreaView style={styles.container}>
       <View style={styles.cameraPreview}>
         {!isIosSimulator && isCameraOn ? (
-          <VideoPreviewView style={styles.cameraPreviewView} />
+          <VideoPreviewView
+            style={styles.cameraPreviewView}
+            mirrorVideo={mirrorPreview}
```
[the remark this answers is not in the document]
karkakol: User can mirror:
The reason why we mirror when swichig from front to back camera etc is because capturer gives mirrored preview. 
That is why i added is front and why there is xor. Few comments before i explained it. And now in our code we only modify `isFrontCammera`. Mirror property stays the same on it can be only modified by user

graszka22 (approved):

PR #34, opened 2024-07-23
Fix notifee on ios

graszka22:
Fix notifee on ios by using patch-package instead of a pr link


graszka22:
Notifee monorepo has a script that copies NotifyCore from the main folder to packages/react-native folder. It's run when setting up the repo / building the package. Initially NotifyCore is not present in packages/react-native folder.
When building it with postinstall script I didn't copy that folder (it wasn't even downloaded as I downloaded only packages/react-native) so ios complained files were missing.
With patch-package the NotifyCore folder is already in the package and I just modify the source and dist folder with patch with changes from the PR.

karkakol (approved):
👍

PR #35, opened 2024-07-23
[FCE-312] Handle auth errors

graszka22:
This PR adds error handling of auth errors like invalid/expired token etc. 
The commands queue no longer ends jobs by throwing exceptions but ends them gracefully.
It also makes errors on RN Android nicer (all possible errors with error messages are specified in one file).

### DIFF
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
@@ -1,28 +1,31 @@
 package com.fishjamcloud.client
 
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.Job
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
 
 internal class CommandsQueue {
   private var commandsQueue: ArrayDeque<Command> = ArrayDeque()
   var clientState: ClientState = ClientState.CREATED
 
-  fun addCommand(command: Command): Job {
+  suspend fun executeCommand(command: Command) {
     commandsQueue.add(command)
-    if (commandsQueue.size == 1) {
-      command.execute()
+    suspendCoroutine { cont ->
+      run {
+        command.continuation = cont
+        if (commandsQueue.size == 1) {
+          command.execute()
+        }
+      }
     }
-    return command.job
   }
 
   fun finishCommand() {
     val command = commandsQueue.first()
-    val job = command.job
     commandsQueue.removeFirst()
     if (command.clientStateAfterCommand != null) {
       clientState = command.clientStateAfterCommand
     }
-    job.complete()
+    command.continuation?.resume(Unit)
     // TODO: make it iterative, not recursive?
     if (commandsQueue.isNotEmpty()) {
       commandsQueue.first().execute()
@@ -41,9 +44,20 @@ internal class CommandsQueue {
     }
   }
 
-  fun clear(cause: String) {
+  fun onDisconnected() {
     clientState = ClientState.CREATED
-    commandsQueue.forEach { command -> command.job.cancel(CancellationException(cause)) }
+    commandsQueue
+      .filter { it.isConnectionCommand() }
+      .forEach { command -> command.continuation?.resume(Unit) }
+    commandsQueue.removeAll { it.isConnectionCommand() }
+    if (commandsQueue.isNotEmpty()) {
+      commandsQueue.first().execute()
+    }
+  }
+
+  fun clear() {
+    clientState = ClientState.CREATED
+    commandsQueue.forEach { command -> command.continuation?.resume(Unit) }
     commandsQueue.clear()
   }
 }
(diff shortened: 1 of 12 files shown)


karkakol:
Update description

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -102,6 +103,18 @@ internal class FishjamClientInternal(
...
 
+        override fun onClosing(
+          webSocket: WebSocket,
+          code: Int,
+          reason: String
+        ) {
+          if (AuthError.isAuthError(reason)) {
+            listener.onAuthError(AuthError.fromString(reason))
+          }
+          commandsQueue.clear("Websocket was closed, reason: $reason")
```
karkakol, line 114: Maybe instead of fixed string lets use enum with code, reason and text message.

karkakol (commented):
Response to my comments

mironiasty (commented):
Can you update PR description? 🥺

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
graszka22: `commandsQueue.clear` no longer throws any exceptions, just ends jobs. Errors in android-client are handled by callbacks like `onSocketError`, `onAuthError` etc.
In rn-client on Android side I've added errors as classes (inspired by expo).
We could try something like this also on iOS side but with ios refactor ongoing it would be a mess.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -142,32 +154,36 @@ internal class FishjamClientInternal(
...
-          OkHttpClient().newWebSocket(
-            request,
-            websocketListener
-          )
-
-        this.webSocket = webSocket
-      }
-    )
+    coroutineScope.launch {
+      commandsQueue.executeCommand(
```
mironiasty, line 163: I find naming here a little... confusing? Should this method be still called `addCommand` as it is not guarantee to always execute passed command?
I guess something like `addAndMaybeExecuteCommand` would be more correct - but it looks strange :p

mironiasty (approved):
One comment about naming. Rest seems to be ok

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
```diff
@@ -1,28 +1,31 @@
...
-import kotlinx.coroutines.Job
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
 
 internal class CommandsQueue {
   private var commandsQueue: ArrayDeque<Command> = ArrayDeque()
   var clientState: ClientState = ClientState.CREATED
 
-  fun addCommand(command: Command): Job {
+  suspend fun executeCommand(command: Command) {
```
karkakol, line 10: I do not like name execute command, because it does not always execute command here. I think that `addCommand` is a better name.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
```diff
@@ -1,28 +1,31 @@
...
     commandsQueue.add(command)
-    if (commandsQueue.size == 1) {
-      command.execute()
+    suspendCoroutine { cont ->
```
karkakol, line 12: why this must be a suspendCorutine?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
```diff
@@ -41,9 +44,20 @@ internal class CommandsQueue {
...
+  fun onDisconnected() {
     clientState = ClientState.CREATED
-    commandsQueue.forEach { command -> command.job.cancel(CancellationException(cause)) }
+    commandsQueue
+      .filter { it.isConnectionCommand() }
+      .forEach { command -> command.continuation?.resume(Unit) }
+    commandsQueue.removeAll { it.isConnectionCommand() }
+    if (commandsQueue.isNotEmpty()) {
+      commandsQueue.first().execute()
+    }
```
karkakol, line 55: Can you explain why all connection commands are removed and then we execure remaining commend? Also what if there is more than one command to execute?

graszka22: we can add/remove tracks when there is no connection, it would be weird to silently not add track because the connection was broken and we don't like throwing errors now
we can't however for example join 

if there is more than one command it will be executed automatically by the queue when the first command finishes

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/CommandsQueue.kt
graszka22: so we can await for finishing the command

karkakol (approved):
👍

PR #54, opened 2024-08-02
[NO-TICKET] Add simple styling for docs

mironiasty:
## Description

Sync documentation styling with `web-client-sdk` and `js-server-sdk`

## Motivation and Context

Unify project configuration

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -23,7 +23,7 @@ concurrency:
 
 jobs:
   # Single deploy job since we're just deploying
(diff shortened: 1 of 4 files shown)


graszka22 (approved):
I think we should have more than a generated documentation soon, especially since we now focus on only react-native sdk. We used generated documentation to have any documentation since we didn't have enough time to maintain proper docs for 6 libraries.

mironiasty:
Proper documentation is no our roadmap. This is just placeholder for now.

PR #55, opened 2024-08-05
[FCE-337] Simplify useMicrophone usage

mironiasty:
## Description

Describe your changes in detail

## Motivation and Context

Why is this change required? What problem does it solve? If it fixes an open
issue, please link to the issue here.

## How has this been tested?

X.X.X)

## Types of changes

      not work as expected)

## Checklist:


examples/video-chat/.env.example
```diff
@@ -1,2 +1,2 @@
-FISHJAM_URL=Your Fishjam URL
+EXPO_PUBLIC_ROOM_MANAGER_URL=Your Fishjam URL
```
mironiasty, line 1: When using expo, we need to use `EXPO_PUBLIC_` prefix

examples/video-chat/hooks/useForegroundService.ts
```diff
@@ -43,17 +38,10 @@ async function startForegroundService() {
...
-
-  const joinRoom = useCallback(async () => {
-    await startForegroundService();
-
-    await startMicrophone({
-      audioTrackMetadata: { active: isMicrophoneAvailable, type: 'audio' },
-      microphoneEnabled: isMicrophoneAvailable,
-    });
-  }, [isMicrophoneAvailable, startMicrophone]);
+export function useForegroundService() {
```
mironiasty, line 41: There will be soon new PR that will update this hook, to contain all logic related to notifee

packages/react-native-client/src/common/state.ts
```diff
@@ -0,0 +1,18 @@
+// hopefully temporary file with global state
```
mironiasty, line 1: I think in long term, we should move it to native code 🤔

packages/react-native-client/src/common/state.ts
```diff
@@ -0,0 +1,18 @@
...
+export function setConnectionStatus(connected: boolean) {
+  isClientConnected = connected;
+}
+
+export function isMicrophoneSetToOn() {
+  return isMicrophoneEnabled;
+}
+export function setMicrophoneStatus(connected: boolean) {
+  isMicrophoneEnabled = connected;
+}
```
mironiasty, line 18: _(I'm open for better naming here)_

graszka22, line 18: Why do we need this? User already has `isMicrophoneOn` in `useMicrophone` and it comes from the native side.

mironiasty: I wanted to make `useMicrophone` usable even before user starts a call. So there is no need to manage manual `setState` for microphone on preview screen. But native `isMicrophoneOn` can be used only if microphone is initialised (after call has started). So I made this hack for now, to make it possible. In future I would like to keep this state on native part - for now I avoided doing that as you and Karol are refactoring native code.

graszka22: You can use `useMicrophone` before the call (just like `useCamera` and `startCamera` btw) and we keep microphone state before the call. At least on Android and probably after Karol's refactor on iOS (that's why currently camera is kinda broken on iOS - the refactor enables camera to be started without a call and we're doing it now in example app). So I'd avoid doing temporary stuff like this just to delete it in two days.

mironiasty: I've tried to do that, but if microphone is started in Android FishjamClient before joining room - warning is thrown that is it not possible to start microphone (AFAIR it because starting microphone is also starting new track, which can be done only if peer connection is established)

packages/react-native-client/src/common/state.ts
karkakol, line 1: Imo this file should not exist. State should be fully placed on native side and only requested by react-native code.
It is a good idea to provide native connection status probably as enum/string.

karkakol (commented):
Mixed fillings about this PR. I think that those changes should be done on native side. When i finish my changes microphone would be accesible before connection on ios

packages/react-native-client/src/common/state.ts
graszka22: There was a warning like this before my refactor, but now it shouldn't be, are you sure about this? 
Also you can start a new local track before establishing peer connection and we're currently doing that with camera.

graszka22 (approved):

PR #57, opened 2024-08-06
[FCE-313] Add reconnection feature

graszka22:
This PR adds a feature for automatic reconnection. If there is no internet the sdk will attempt to connect again with exponential backoff. User can configure the amount of attempts and the delay. User can disable this by setting max attempts to 0. 
When reconnecting we must "recreate" local tracks (if we don't do this, we'll crash the backend with duplicated tracks error). But to avoid opening up the camera and all this mess with screencast permissions etc. we're reusing track sources, so the experience is smoother.
Also app developer can use a hook to keep track of reconnection status and inform app users what's happening.

### DIFF
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClient.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClient.kt
@@ -10,6 +10,7 @@ import com.fishjamcloud.client.models.EncoderOptions
 import com.fishjamcloud.client.models.Metadata
 import com.fishjamcloud.client.models.Peer
 import com.fishjamcloud.client.models.RTCStats
+import com.fishjamcloud.client.models.ReconnectConfig
 import com.fishjamcloud.client.models.TrackBandwidthLimit
 import com.fishjamcloud.client.models.TrackEncoding
 import com.fishjamcloud.client.models.VideoParameters
@@ -21,7 +22,9 @@ import org.webrtc.Logging
 
 data class Config(
   val websocketUrl: String,
-  val token: String
+  val token: String,
+  val peerMetadata: Metadata,
+  val reconnectConfig: ReconnectConfig
 )
 
 class FishjamClient(
@@ -50,17 +53,6 @@ class FishjamClient(
    */
   fun leave() {
     client.leave()
-  }
-
-  /**
-   * Tries to join the room. If user is accepted then {@link FishjamClient.onConnected} will be called.
-   * In other case {@link FishjamClient.onConnectError} is invoked.
-   *
-   * @param peerMetadata - Any information that other peers will receive in onPeerJoined
-   * after accepting this peer
-   */
-  fun join(peerMetadata: Metadata = emptyMap()) {
-    client.join(peerMetadata)
   }
 
   /**
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
@@ -31,7 +31,6 @@ import com.fishjamcloud.client.webrtc.PeerConnectionListener
 import com.fishjamcloud.client.webrtc.PeerConnectionManager
 import com.fishjamcloud.client.webrtc.RTCEngineCommunication
 import com.fishjamcloud.client.webrtc.RTCEngineListener
-import com.github.ajalt.timberkt.BuildConfig
 import fishjam.PeerNotifications
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -62,10 +61,15 @@ internal class FishjamClientInternal(
   private var webSocket: WebSocket? = null
 
   private var localEndpoint: Endpoint = Endpoint(id = "", type = EndpointType.WEBRTC)
+  private var prevTracks = mutableListOf<Track>()
+  private var videoParameters = VideoParameters.presetFHD169
   private var remoteEndpoints: MutableMap<String, Endpoint> = mutableMapOf()
 
   private val coroutineScope: CoroutineScope =
     ClosableCoroutineScope(SupervisorJob() + Dispatchers.Default)
+  private var config: Config? = null
+
+  private lateinit var reconnectionManager: ReconnectionManager
 
   init {
     if (BuildConfig.DEBUG) {
@@ -90,8 +94,23 @@ internal class FishjamClientInternal(
         ?.firstOrNull { track -> track.getRTCEngineId() == trackId }
 
   fun connect(config: Config) {
+    this.config = config
+    this.reconnectionManager =
+      ReconnectionManager(config.reconnectConfig) {
+        reconnect(config)
+      }
     peerConnectionManager.addListener(this)
     rtcEngineCommunication.addListener(this)
+    reconnectionManager.addListener(listener)
+    setupWebsocket(config)
+  }
+
+  private fun reconnect(config: Config) {
+    recreateTracks()
+    setupWebsocket(config)
+  }
+
+  private fun setupWebsocket(config: Config) {
     val websocketListener =
       object : WebSocketListener() {
         override fun onClosed(
@@ -110,8 +129,9 @@ internal class FishjamClientInternal(
         ) {
           if (AuthError.isAuthError(reason)) {
             listener.onAuthError(AuthError.fromString(reason))
+          } else {
+            webSocket.close(code, reason)
           }
-          webSocket.close(code, reason)
         }
 
         override fun onMessage(
@@ -123,6 +143,7 @@ internal class FishjamClientInternal(
             if (peerMessage.hasAuthenticated()) {
               listener.onAuthSuccess()
               commandsQueue.finishCommand()
+              join()
             } else if (peerMessage.hasMediaEvent()) {
               receiveEvent(peerMessage.mediaEvent.data)
             } else {
@@ -156,6 +177,10 @@ internal class FishjamClientInternal(
         ) {
           listener.onSocketError(t)
           commandsQueue.onDisconnected()
+          coroutineScope.launch {
+            onDisconnected()
+            reconnectionManager.onDisconnected()
+          }
         }
       }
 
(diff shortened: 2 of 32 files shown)


packages/react-native-client/ios/RNFishjamClient.swift
```diff
@@ -268,7 +268,6 @@ class RNFishjamClient: FishjamClientListener {
     }
 
     func onAuthSuccess() {
-        joinRoom()
     }
```
mironiasty, line 271: Can this method be removed too?

packages/ios-client/Sources/MembraneRTC/MembraneRTC.swift
```diff
@@ -715,6 +717,51 @@ public class MembraneRTC: MulticastDelegate<MembraneRTCDelegate>, ObservableObje
             break
         }
     }
+
+    func onDisconnected() {
+        DispatchQueue.webRTC.sync {
+            peerConnectionManager.close()
+            prevLocalTracks = Array((prevLocalTracks + localTracks))
```
mironiasty, line 724: What would be the condition that `onDisconnect` is called with non empty `prevLocalTracks`?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -90,8 +94,23 @@ internal class FishjamClientInternal(
...
+    reconnectionManager.addListener(listener)
+    setupWebsocket(config)
+  }
+
+  private fun reconnect(config: Config) {
+    recreateTracks()
+    setupWebsocket(config)
+  }
+
+  private fun setupWebsocket(config: Config) {
```
mironiasty, line 113: nit: in most places in code we use `webSocket`
```suggestion
  private fun setupWebSocket(config: Config) {
```

packages/react-native-client/src/common/client.ts
```diff
@@ -1,12 +1,21 @@
 import { Metadata } from '../types';
 import RNFishjamClientModule from '../RNFishjamClientModule';
 
+export type Config = {
+  reconnectConfig?: {
+    maxAttempts?: number;
+    initialDelayMs?: number;
+    delayMs?: number;
```
mironiasty, line 8: Just an idea, I'm not sure about it:
can we make these values mandatory in native code, but define default values in RN SDK? This way there will be single place with default values (right now there are 3+?), but they will not be mandatory for RN SDK users (but they will be mandatory for native SDKs users...)

graszka22: Record fields in expo modules must have a default value: https://docs.expo.dev/modules/module-api/#records
And making the values mandatory for native SDK users would be mean for them. I'd leave it as it is, if we'd like to change this we'd use search option anyway.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    if (reconnectionStatus == ReconnectionStatus.WAITING) {
+      return
+    }
+
+    if (reconnectionStatus != ReconnectionStatus.RECONNECTING) {
+      reconnectionStatus = ReconnectionStatus.RECONNECTING
+      listeners.forEach { it.onReconnectionStarted() }
+    }
+    val delay = reconnectConfig.initialDelayMs + reconnectAttempts * reconnectConfig.delayMs
```
karkakol, line 52: lets put initialDelay = delay = 100, reconnect attempts = 5
Now process looks of connection looks like:
Failure occurs
1st attempt to reconnect, connect is called after 100 + 0 * 100 
Failure occurs again
2 attempt to reconnect, connect is called after 100 + 1 * 100
...
5 attempt to reconnect, connect is called after 100 + 4 * 100
total time of waiting for reconnecting is 1500 
Description says it is exponential backoff, while it is linear because we increase delay by delayMs. 
To make it exponential backoff it should be sth like:
`val delay = min(maxDelay, pow(2, reconnectAttempts) * delay)` 
I think that it should be documented somewhere how delay is calculated before merging it to main, and it should be decides is it some linear delay or exponential backoff.

graszka22: https://github.com/fishjam-cloud/web-client-sdk/blame/main/packages/ts-client/src/reconnection.ts#L144 I followed the web client
So I'll change the name to "linear backoff"?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -62,10 +61,15 @@ internal class FishjamClientInternal(
   private var webSocket: WebSocket? = null
 
   private var localEndpoint: Endpoint = Endpoint(id = "", type = EndpointType.WEBRTC)
+  private var prevTracks = mutableListOf<Track>()
```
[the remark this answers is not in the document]
graszka22: because we only need tracks, not the whole endpoint, so we keep track and name it "prevTracks" 🤷‍♀️

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -0,0 +1,78 @@
+package com.fishjamcloud.client
+
+import com.fishjamcloud.client.models.ReconnectConfig
+import timber.log.Timber
+import java.util.Timer
+import kotlin.concurrent.schedule
+
+interface ReconnectionManagerListener {
```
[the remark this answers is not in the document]
graszka22: it's for the native sdk user / rn client and it's used here: https://github.com/fishjam-cloud/mobile-client-sdk/pull/57/files#diff-25d9b003e6181a81e4a04e7acadbea1cf4e6887f27bc20347a11dbb475948869R3 and in rn client

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -0,0 +1,78 @@
...
+  RECONNECTING,
+  ERROR,
+  WAITING
+}
+
+internal class ReconnectionManager(
+  private var reconnectConfig: ReconnectConfig = ReconnectConfig(),
+  private val connect: () -> Unit
+) {
+  private val listeners = mutableListOf<ReconnectionManagerListener>()
```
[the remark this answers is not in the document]
graszka22: No particular reason, just that we have a list of listeners everywhere now, because it's easier to set up tests that way. But I can change it to a variable if there is an issue with a list.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClient.kt
```diff
@@ -21,7 +22,9 @@ import org.webrtc.Logging
 
 data class Config(
```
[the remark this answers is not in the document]
graszka22: I'll rename it to `ConnectConfig`,just like in rn client.
WebSocketConfig wouldn't be a good name, because it's more than config for web socket (for example peer metadata)

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -175,12 +198,21 @@ internal class FishjamClientInternal(
     }
   }
 
-  fun join(peerMetadata: Metadata = emptyMap()) {
+  private suspend fun onDisconnected() {
```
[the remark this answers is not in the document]
graszka22: Because it would collide with cleaning up in the `fun leave()` when we leave and clean up everything completely. Here we don't want to clean up everything, just "prepare" for reconnection

packages/ios-client/Sources/FishjamClient/ReconnectionManager.swift
```diff
@@ -0,0 +1,81 @@
...
+    private let connect: () -> Void
+    private var listener: ReconnectionManagerListener
+
+    init(reconnectConfig: ReconnectConfig, connect: @escaping () -> Void, listener: ReconnectionManagerListener) {
+        self.reconnectConfig = reconnectConfig
+        self.connect = connect
+        self.listener = listener
+    }
+
+    func onDisconnected() {
```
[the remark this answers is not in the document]
karkakol: in my opinion this function should be renamed. It is beter for function name to describe what it does that when it is called. Not always this rule is obligatory but in this case i think so

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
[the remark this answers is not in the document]
karkakol: this function name does not suggest that its preparing for reconnections. Now we have in several files and clasess functions like on disconeccted, disconnect, leave etc. Almost everyone of those functions does somthing different and it would be better to somewhere specify use of this function. It does not have to be another function name but at least comment with description.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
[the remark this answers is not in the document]
karkakol: let it be

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
karkakol: What are pros of linear compared to exponential?
I heard that exponential backoff is better. Mayby lets implement it? 
Not sure which one is better but in most cases it is used.

graszka22: I also don't know which one is better, probably it doesn't matter because user will restart the app after ~10 seconds anyway, so I made it to be the same as web for consistency

packages/ios-client/Sources/FishjamClient/ReconnectionManager.swift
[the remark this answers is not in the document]
graszka22: I think this name is ok, this function is supposed to inform the reconnection manager that there was a disconnection and it should do its stuff, whatever it is, we don't care
similarly `onReconnected` informs the reconnection manager that we're connected again and it should do its stuff

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -62,10 +61,15 @@ internal class FishjamClientInternal(
   private var webSocket: WebSocket? = null
 
   private var localEndpoint: Endpoint = Endpoint(id = "", type = EndpointType.WEBRTC)
+  private var prevTracks = mutableListOf<Track>()
+  private var videoParameters = VideoParameters.presetFHD169
```
mironiasty, line 65: Where is this variable used? 🤔

mironiasty (approved):
One question, other stuff seems to be ok 👍

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
graszka22: removed 👍

PR #58, opened 2024-08-06
[FCE-212] Rename references from "endpoint" to "peer" in RN-SDK API

graszka22:
React Native SDK exports functions/types that reference both Peers and Endpoints. From an end user perspective, they are basically the same thing. So, let's rename/remove references to them from public API.


mironiasty (approved):
LGTM

karkakol (approved):
Change pr name to sth like:
`Rename references from endpoint to peer`

PR #68, opened 2024-08-12
Simplify useMicrophone more

graszka22:
- merge main
- remove `startMicrophone`, just leave `useMicrophone` and `toggleMicrophone` - if the local audio track is not yet created - just create it
- you can toggle microphone any time - before connecting or after connecting without warnings
- remove `state.ts` - whether the microphone is on or off is kept on the native side and returned within `useMicrophone`


mironiasty (approved):
LGTM 👍 
(I didn't test it yet, but code itself looks fine)

PR #72, opened 2024-08-21
[FCE-378] Fix Room Manager support

mironiasty:
## Description

Add URL suffix manually, as Room Manager will no longer return in
This is related to this change: https://github.com/fishjam-cloud/js-server-sdk/pull/25

## Motivation and Context

To make example app work with latest Room Manager

### DIFF
--- a/examples/video-chat/screens/ConnectWithRoomManagerScreen.tsx
+++ b/examples/video-chat/screens/ConnectWithRoomManagerScreen.tsx
@@ -28,9 +28,15 @@ async function getFishjamServer(
   roomName: string,
   userName: string,
 ) {
-  const url = roomManagerUrl.endsWith('/')
-    ? roomManagerUrl
-    : roomManagerUrl + '/';
+  // in case user copied url from admin panel
+  const urlWithoutParams = roomManagerUrl.replace(
+    '/*roomName*/users/*username*',
+    '',
+  );
+  // trim slash from end
+  const url = urlWithoutParams.endsWith('/')
+    ? urlWithoutParams
+    : urlWithoutParams + '/';
   const response = await fetch(
     `${url}${roomName.trim()}/users/${userName.trim()}`,
   );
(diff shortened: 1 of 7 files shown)


examples/video-chat/screens/ConnectWithRoomManagerScreen.tsx
```diff
@@ -28,9 +28,15 @@ async function getFishjamServer(
...
   userName: string,
 ) {
-  const url = roomManagerUrl.endsWith('/')
-    ? roomManagerUrl
-    : roomManagerUrl + '/';
+  // in case user copied url from admin panel
+  const urlWithoutParams = roomManagerUrl.replace(
+    '/*roomName*/users/*username*',
+    '',
+  );
```
mironiasty, line 35: This line is added just to simplify example app usage. You can just copy URL from Cloud Panel

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClient.kt
```diff
@@ -20,12 +20,19 @@ import com.fishjamcloud.client.webrtc.PeerConnectionManager
...
+  val reconnectConfig: ReconnectConfig,
+  val peerMetadata: Metadata
+) {
+  public constructor(
+    url: String,
+    token: String,
+    peerMetadata: Metadata,
+    reconnectConfig: ReconnectConfig
+  ) : this(websocketUrl = "$url/socket/peer/websocket", token = token, peerMetadata = peerMetadata, reconnectConfig = reconnectConfig) {}
+}
```
mironiasty, line 35: What could be better way to handle this logic in Kotlin? 🤔

karkakol:
```
data class ConnectConfig(
  private _websocketUrl: String,
  val token: String,
  val peerMetadata: Metadata,
  val reconnectConfig: ReconnectConfig
){
 val websocketUrl : String
    get() = "$_websocketUrl/socket/peer/websocket"
}
```

Something like this should work ;)

PR #78, opened 2024-08-22
[FCE-400] Improve camera grid view

mironiasty:
## Description

This PR simplifies how camera grid works in mobile app. 
- removing animations (they were breaking styling, and we don't really need them in demo app now
- show user names in each video tile
- do not change video aspect ratio
- set correct user name when connection starts

## Motivation and Context

This changes are mostly to make debugging and working on future features a little easier.

## How has this been tested?

Manual testing

## Types of changes

      not work as expected)

## Screenshots (if appropriate)
Screenshot 2024-08-22 at 15 06 50

### DIFF
--- a/examples/video-chat/components/VideosGrid.tsx
+++ b/examples/video-chat/components/VideosGrid.tsx
@@ -2,34 +2,68 @@ import {
   Track,
   VideoRendererView,
   Metadata,
+  Peer,
 } from '@fishjam-cloud/react-native-client';
 import React from 'react';
 import { FlatList, StyleSheet, View } from 'react-native';
 
 import { roomScreenLabels } from '../types/ComponentLabels';
 import { BrandColors } from '../utils/Colors';
+import Typo from './Typo';
 
 type Props = {
-  tracks: Track<Metadata>[];
+  tracks: GridTrack[];
+};
+
+type GridTrack = Track<Metadata> & {
+  isLocal: boolean;
+  userName: string | undefined;
 };
 
 const { VIDEO_CELL } = roomScreenLabels;
 
+export function parsePeersToTracks(
+  peers: Peer<Metadata, Metadata, Metadata>[],
+): GridTrack[] {
+  return peers
+    .sort((peer) => (peer.isLocal ? -1 : 1))
+    .flatMap((peer) =>
+      peer.tracks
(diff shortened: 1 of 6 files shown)


examples/video-chat/screens/ConnectWithTokenScreen.tsx
```diff
@@ -68,13 +68,13 @@ const ConnectScreen = ({ navigation }: Props) => {
           />
           <TextInput
             onChangeText={onChangeFishjamUrl}
-            value={fishjamUrl}
+            defaultValue={fishjamUrl}
```
mironiasty, line 71: This is for debugging -> I can easily hardcode room/user name in code so I don't have to type it with each app restart

examples/video-chat/components/VideosGrid.tsx
```diff
@@ -1,119 +1,100 @@
...
-            trackId={v.id}
-            entering={FadeInDown.duration(200)}
-            layout={LinearTransition.duration(150)}
-            style={styles.animatedView}
+          style={[
+            styles.video,
+            {
+              backgroundColor: track.isLocal
+                ? BrandColors.yellow100
+                : BrandColors.darkBlue60,
```
mironiasty, line 55: This colors can be considered ugly. I'm open for proposals here.

karkakol (approved):
👍

examples/video-chat/screens/ConnectWithTokenScreen.tsx
karkakol: It can be a feature to add for "FIT" property.
Make it that way so in background is this screen as "FILL" and added blur to it. Many ios apps use such patter.

examples/video-chat/components/VideosGrid.tsx
karkakol: It can be a feature for "FIT" property.
My proposal is to add background view as "FILL" and then blur it. Many aps use that patter for fitting videos/images. What do you think about it?

mironiasty: For now I would prefer to keep this app simple, to make sure that our core functionalities works. Bluring/animations - it is nice, but we don't have time to support it now :(

PR #80, opened 2024-08-22
Fix room manager audio tracks

graszka22:
To reproduce:
- run room manager
- add some peers with video AND audio tracks
- connect mobile to the room
- some tracks will be black squares

What happened? The old videoroom was sending track metadata which contained "type" value set to "audio" for audio track. The new videoroom doesn't send any metadata. So the example app treated audio tracks like video track.

Fixed by checking the type of the track, not metadata.

I'm not sure if this is the bug that @mironiasty have experienced but after fixing that everything works ok for me - I can see all the video tracks, no black squares and I can turn on/off some of the tracks.


mironiasty (approved):
Great fix! Thanks :)

PR #81, opened 2024-08-23
[FCE-402] Simplify start camera api

mironiasty:
## Description

Remove possibility to set Track Metadata when enabling camera
Simplify simulcast configuration (as SDK will now only enable/disable this config - and values will be set automatically)

## Motivation and Context

Make our API simpler

## How has this been tested?

It was tested on example app

## Types of changes

      not work as expected)

## Checklist:


examples/video-chat/components/VideosGrid.tsx
```diff
@@ -35,8 +35,7 @@ export function parsePeersToTracks(
           userName: peer.metadata?.name,
         }))
         .filter(
-          (track) =>
-            track.metadata.type !== 'audio' && (track.metadata.active ?? true),
+          (track) => track.type === 'Video' && (track.metadata.active ?? true),
```
mironiasty, line 38: This is same fix as applied in #80 by @graszka22

packages/react-native-client/src/hooks/useCamera.ts
```diff
@@ -45,41 +46,96 @@ export type CameraConfig<MetadataType extends Metadata> = {
    * whether to flip the dimensions of the video, that is whether to film in vertical orientation.
    * @default `true`
    */
-  flipVideo?: boolean;
+  flipVideo?: boolean /**
```
graszka22, line 49: weird formatting

examples/video-chat/screens/PreviewScreen/PreviewScreen.tsx
```diff
@@ -87,17 +87,9 @@ function PreviewScreen({
...
-          enabled: true,
-          activeEncodings:
-            // iOS has a limit of 3 hardware encoders
-            // 3 simulcast layers + 1 screencast layer = 4, which is too much
-            // so we limit simulcast layers to 2
-            Platform.OS === 'android' ? ['l', 'm', 'h'] : ['l', 'h'],
-        },
+        simulcastEnabled: true,
         quality: 'HD169',
         maxBandwidth: { l: 150, m: 500, h: 1500 },
```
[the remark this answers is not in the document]
graszka22: I think those simulcast settings are a leftover from the times this was a client for RTCEngine. Back then, it was the user who was responsible for setting up backend and they could do whatever they wanted. If they had much resources they could allow for high quality video. If not - they could set a low quality video. 
Also if they use screencast - they must limit layers to ios to 2. However if they don't use screencast in the app at all they can have 3 layers of simulcast.
Right now this is a client for Fishjam Cloud, and we're responsible for the backend and scalability so managing simulcast should be our responsibility? I'm not even sure if Fishjam Cloud supports simulcast at all - I've heard it doesn't and maybe will be added in the future? In this case even setting `simulcastEnabled: true` has no sense? 
We can also think about `maxBandwidthLimit` - if we should allow user to set this.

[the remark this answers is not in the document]
mironiasty: AFAIK we support simulcast for now. But we might drop this support with introduction of WebRTC. As one of next steps, we should better test how simulcast works on mobile now, and how other video quality settings are working. But it is something to be done in near future.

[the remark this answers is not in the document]
mironiasty: Thanks for explaining how it used to work in past

packages/react-native-client/src/hooks/useCamera.ts
mironiasty: Indeed, but should be better now

packages/react-native-client/src/hooks/useCamera.ts
```diff
@@ -45,41 +46,96 @@ export type CameraConfig<MetadataType extends Metadata> = {
...
-  videoTrackMetadata?: MetadataType;
+  captureDeviceId?: string;
+};
+
+export type CameraConfig = CameraConfigBase & {
   /**
    *  SimulcastConfig of a video track. By default simulcast is disabled.
    */
-  simulcastConfig?: SimulcastConfig;
+  simulcastEnabled?: boolean;
```
karkakol, line 69: Is comment above outdated or should it be updated?

karkakol (commented):
Please answer to my comment. Then approve.

packages/react-native-client/src/hooks/useCamera.ts
mironiasty: I missed that comment has different prop name. Anyway, I've updated it!

graszka22 (approved):

PR #82, opened 2024-08-23
[FCE-390] Remove metadata from track info

mironiasty:
## Description

Remove metadata from tracks returned from SDK

## Motivation and Context

These values should be opaque to user.

### DIFF
--- a/packages/react-native-client/src/hooks/usePeers.ts
+++ b/packages/react-native-client/src/hooks/usePeers.ts
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react';
 
-import { Metadata, TrackEncoding } from '../types';
+import { Metadata, TrackEncoding, TrackMetadata } from '../types';
 import RNFishjamClientModule from '../RNFishjamClientModule';
 import { ReceivableEvents, eventEmitter } from '../common/eventEmitter';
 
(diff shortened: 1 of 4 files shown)


packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -14,19 +14,26 @@ export type TrackType = 'Audio' | 'Video';
...
+  type: 'Audio';
+  vadStatus: VadStatus | undefined;
+};
+
+export type VideoTrack = TrackBase & {
+  type: 'Video';
   // Encoding that is currently received. Only present for remote tracks.
   encoding: TrackEncoding | null;
-  // Information about simulcast, if null simulcast is not enabled
-  simulcastConfig: SimulcastConfig | null;
```
mironiasty, line 25: It seems that native code never set these values 🤔

packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -68,40 +63,38 @@ export type Peer<
...
 
+function addIsActiveToTracks<MetadataType extends Metadata>(
+  peers: ReadonlyArray<Peer<MetadataType>>,
+): Peer<MetadataType>[] {
+  return peers.map((peer) => ({
+    ...peer,
+    tracks: peer.tracks.map((track) => ({
+      ...track,
+      isActive:
+        (track as { metadata?: TrackMetadata })?.metadata?.active ?? true,
```
graszka22, line 77: does the web sdk also pass `active` and `type` in metadata?

mironiasty: Web SDK often rely on just removing track, so it is not always needed. But @kamil-stasiak is working on changes to also remove metadata handling from part of web SDK. So at some point, we could sync mobile and web approach

karkakol (approved):
👍

PR #83, opened 2024-08-23
[FCE-402] Use branded type as camera id

mironiasty:
## Description

Branded types are nice tool for restricting type of values that can be passed into functions.
This change converts camera id into such type. This way, when switching camera, developer won't be able to pass arbitrary string. It will be possible to pass only id we are allowing to use.

## Motivation and Context

Add better restriction to API

## Types of changes

      not work as expected)


graszka22 (approved):
TIL 👀

karkakol (approved):
Interesting and useful thing 👍

PR #84, opened 2024-08-23
[FCE-408] Update screencast metadata

mironiasty:
## Description

Remove screencast metadata from public API. Let's use hardcoded version inside library.

## Motivation and Context

Simplify API


karkakol (approved):
👍

PR #95, opened 2024-09-05
[FCE-478] Rename mobile test app

MiloszFilimowski:
## Description

Renamed example app from `video-chat` to to `fishjam-chat`

## Motivation and Context

Use the same name for the example as in the [web-client-sdk](https://github.com/fishjam-cloud/web-client-sdk)

## How has this been tested?

- Built and run both Android and iOS
- Run e2e tests. Tests failed locally, but built and run successfully. Created new [issue to fix](https://linear.app/swmansion/issue/FCE-481/fix-local-e2e-example-tests)

## Types of changes

      not work as expected)

## Checklist:

- [x] I have updated the documentation accordingly.


examples/fishjam-chat/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "video-chat-example",
+  "name": "fishjam-chat-example",
```
mironiasty, line 2: Let's use similar approach as in `web-client-sdk` and have:
```suggestion
  "name": "@fishjam-example/fishjam-chat",
```

mironiasty (approved):
Maybe let's do it in separate PR - but I think it would be also worth to update app name from FishjamExample to FishjamChat
@karkakol WDYT?

MiloszFilimowski:
> Maybe let's do it in separate PR - but I think it would be also worth to update app name from FishjamExample to FishjamChat @karkakol WDYT?

I thought about this but decided it was too much effort compared to what's gained, as native renames usually are difficult to perform.

My proposition is after we bump the Expo version, we could recreate the example app and migrate the UI. It will be easier to start fresh and we would have a clean example project with proper names everywhere.

examples/fishjam-chat/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "video-chat-example",
+  "name": "@fishjam-example/fishjam-chat",
```
karkakol, line 2: Would not it be better to just name it `fishjam-chat`?
It is shorter, easier to write and looks more natural imho.

karkakol (approved):
image

examples/fishjam-chat/package.json
MiloszFilimowski: As @mironiasty has suggested, this will now be consistent with the web SDK.

PR #96, opened 2024-09-05
Accept Fishjam's path as Room Manager's

PiotrWodecki:
## Description

This change provides better heuristics for fixing copy-pasted connection URLs.

## Motivation and Context

There are three places that the user can get the connection URL:

- admin panel main page
- app view
- self-crafted from the app view

URL entry now allows for the app-view-formatted URL, by appending `/room-manager`. We also trim the string as a bonus.

Without this, the user would get a very misleading `Not Found` error from the Fishjam instance.

## How has this been tested?

Manual on both Android and iOS emulator/simulator.

## Types of changes

      not work as expected)

## Checklist:


mironiasty (approved):
👍

PR #97, opened 2024-09-05
[FCE-483] Fix deleting remote tracks

karkakol:
## Description

Deleting and adding remote tracks does not worked properly

## How has this been tested?

on IOS

## Types of changes

      not work as expected)

## Checklist:


karkakol:
FCE-483

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
```diff
@@ -439,12 +439,12 @@ internal class FishjamClientInternal: WebSocketDelegate, PeerConnectionListener,
...
         if localEndpoint.id == endpointId { return }
 
-        guard let endpoint = remoteEndpointsMap.removeValue(forKey: endpointId) else {
+        guard let endpoint = remoteEndpointsMap[endpointId] else {
             sdkLogger.error("Failed to process TracksAdded event: Endpoint not found: \(endpointId)")
             return
         }
 
-        var updatedTracks = [String: Track]()
+        var updatedTracks: [String: Track] = endpoint.tracks
```
MiloszFilimowski, line 447: Do you still need type declaration?

karkakol: no
but i do not consider it a bad habbit

packages/ios-client/Sources/FishjamClient/media/Tracks/VideoTrack.swift
```diff
@@ -5,7 +5,7 @@ open class VideoTrack: Track {
         mediaTrack: RTCVideoTrack, endpointId: String, rtcEngineId: String?, metadata: Metadata = Metadata(),
         id: String = UUID().uuidString
     ) {
-        super.init(mediaTrack: mediaTrack, endpointId: endpointId, rtcEngineId: rtcEngineId, metadata: metadata)
+        super.init(mediaTrack: mediaTrack, endpointId: endpointId, rtcEngineId: rtcEngineId, metadata: metadata, id: id)
```
mironiasty, line 8: Would it make sense to do not have `id` as optional param in `Track`?

mironiasty (approved):

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
MiloszFilimowski: Sure! My point is that it doesn't provide any more information and you can always setup your IDE to hint types if you like.

What do you think we would commit to a predefined style guide?

MiloszFilimowski (approved):

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
karkakol: Predefined style guide it is a great idea!
Lets create task for it and apply it the future.

MiloszFilimowski: Great! 
https://linear.app/swmansion/issue/FCE-492/style-guide-for-iosswift

packages/ios-client/Sources/FishjamClient/media/Tracks/VideoTrack.swift
karkakol: I think it would make sense for id to be a required param in initializer and on the other hand it would make sense to be optional param. The reason I made it optional is android shares the same behaviour.

PR #104, opened 2024-09-11
Remove husky from project

mironiasty:
## Description

Remove husky from project. So there will be no commit checks
Also fix lint scripts to catch errors

## Motivation and Context

Commit checks in monorepo are difficult to configure. As in perfect world, precommit action should only check for staged files. But our scripts always run on all files. So it might catch some false positives during development and be annoying for developer.

Let's rely on PR checks for now, and maybe in future we can update our tooling to have better checking

## How has this been tested?

X.X.X)

## Types of changes

      not work as expected)

## Checklist:


karkakol (approved):
👍

PR #105, opened 2024-09-11
[FCE-502] Fix Android crashing on sdkVersion >= 34

MiloszFilimowski:
## Description

The whole issue is related to permissions for `mediaProjection`. Somehow the previous implementation (even though it looked fine) requested the foreground service with `mediaProjection` on launch, without permission first. 

- Added an Expo Module for Android which handles starting a foreground service for the user (required for Android >= 34) 
- Added an Expo Config Plugin to allow conditional enabling of service

**Discussion**: I believe, ideally, to provide the best experience for our users, we should start the service implicitly inside the SDK after requesting screen sharing permissions (if the user has `enableForegroundService` set to true in our plugin). Right now there is a `if == android` in our example which is not ideal and not relevant for most users. 

## Motivation and Context

The previous implementation that used Notifee was removed with the Expo upgrade. 

## How has this been tested?

- Successfully launched Screen sharing on Android 34
- Launched iOS Room Screen with no issues

## Types of changes

      not work as expected)

## Checklist:


scripts/init.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 brew install swift-format ktlint release-it
 yarn
-yarn build
+yarn prepare
```
mironiasty, line 4: Are you sure that his is expected change? with `build` we wanted TS files to be generated for local usage + release. `prepare` should do other stuff 🤔

packages/react-native-client/.eslintrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../../.eslintrc.js');
```
mironiasty, line 1: Curious: why this change?

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/FishjamForegroundService.kt
```diff
@@ -0,0 +1,90 @@
+package org.membraneframework.reactnative
```
mironiasty, line 1: TODO: we should remove membraneframework namespace at some point...

mironiasty (approved):
Looks ok. But it would be good if @graszka22 would also take a look before merge

packages/react-native-client/.eslintrc.js
MiloszFilimowski: `expo prepare` generates this file, but with a wrong path. Adding it manually fixes the issue.

scripts/init.sh
MiloszFilimowski: The thing was that `expo-module build` doesn't build the plugin on its own, causing the example not to work. Additionally, it runs with `--watch` automatically so there is a need for `EXPO_NONINTERACTIVE=1` before every command. 

When you look at the [source code of prepare](https://github.com/expo/expo/blob/main/packages/expo-module-scripts/bin/expo-module-prepare) ` it just runs `expo-module build` with different params to build everything. 

Anyway, I reverted to the previous and updated `package.json` to run only the builds that we needed.

package.json
```diff
@@ -27,7 +27,8 @@
   },
   "scripts": {
     "setup": "sh ./scripts/init.sh",
-    "build": "yarn workspace @fishjam-cloud/react-native-client run tsc",
+    "build": "yarn workspace @fishjam-cloud/react-native-client run build",
+    "prepare": "yarn workspace @fishjam-cloud/react-native-client run prepare",
```
graszka22, line 31: what does yarn build and yarn prepare do and when should we run it? it should be described in README

packages/react-native-client/android/src/main/java/org/membraneframework/reactnative/FishjamForegroundService.kt
MiloszFilimowski: [FCE-528](https://linear.app/swmansion/issue/FCE-528)

graszka22 (commented):
minor comments, the service stuff LGTM 💪

examples/fishjam-chat/hooks/useForegroundService.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  startForegroundService,
+  stopForegroundService,
+} from '@fishjam-cloud/react-native-client';
+import { AndroidForegroundServiceType } from '@fishjam-cloud/react-native-client/build/types';
```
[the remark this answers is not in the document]
MiloszFilimowski: Great catch! I forgot to export it 🤦

package.json
MiloszFilimowski: It just runs `expo-module build` and `expo-module prepare` in react-native-client. 

What would you like to see in the README? It already contains a setup script that uses `yarn build`.
`yarn prepare` is only relevant when releasing.

graszka22: It would be good to see what `yarn prebuild` does and when it's needed. "is only relevant when releasing" - why do we need to run this when releasing? Do we need to run this when first checking out the repo?

PR #110, opened 2024-09-13
[FCE-524] Investigate screen broadcast on iOS after updating EXPO

MiloszFilimowski:
## Description

This is mostly the old plugin, which was updated to work with our new setup. 

## Motivation and Context

Creating iOS native files was broken after upgrading to a new expo version. 

## How has this been tested?

- Successfully run clean prebuild and launch iOS app. 

## Types of changes

## Checklist:


examples/fishjam-chat/app.json
```diff
@@ -49,9 +49,13 @@
...
           "android": {
             "enableForegroundService": true
+          },
+          "ios": {
+            "enableScreensharing": true
           }
         }
-      ]
+      ],
+      ["./plugin/build/development-pods-plugin.js"] // Only relevant for local development of packages/ios-client
```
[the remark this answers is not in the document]
MiloszFilimowski: AFAIK expo uses [JSON5](https://json5.org/) under the hood for app.json so it should be perfectly fine.

README.md
```diff
@@ -23,12 +23,14 @@ room client. To run the app:
...
 3. ```
-   yarn
-   yarn build
-   cd `examples/fishjam-chat`
-   yarn prebuild
-   yarn run android/ios
+      yarn
+      yarn build
+      cd `examples/fishjam-chat`
+      npx expo prebuild --clean
```
mironiasty, line 29: why this change? 🤔

examples/fishjam-chat/package.json
```diff
@@ -8,7 +8,7 @@
     "android": "expo run:android",
     "ios": "expo run:ios",
     "test": "jest",
-    "build": "tsc",
+    "build": "tsc && tsc -p ./plugin",
```
mironiasty, line 11: What does `-p` options does? I can't find it in [official docs](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

examples/fishjam-chat/app.json
mironiasty, line 58: WDHY about having this plugin in package?
(might not be part of release in future)
This way, if we would create another example app, we can easily reuse it

packages/react-native-client/plugin/src/withFishjamIOS.ts
```diff
@@ -0,0 +1,261 @@
+// ios-related code was mostly copied from OneSignal expo plugin: https://github.com/OneSignal/onesignal-expo-plugin/blob/main/onesignal/withOneSignalIos.ts
```
mironiasty, line 1: WDYT about adding a little info in comment somewhere about what this plugin does?

README.md
MiloszFilimowski: What change do you mean? The `npx expo prebuild` or the removed `yarn build` step?

examples/fishjam-chat/package.json
MiloszFilimowski: It's the same as `--project`. I'm surprised it's not mentioned in the docs however it's there when you run `tsc --help`.

examples/fishjam-chat/app.json
MiloszFilimowski: What other project do you mean? Maybe when we have another one let's think about it then? For playground, not releasable projects it's easier to just copy.

packages/react-native-client/plugin/src/withFishjamIOS.ts
MiloszFilimowski: I added some additional comments on what each step does.

README.md
mironiasty: changing `yarn prebuild` to `npx expo prebuild`

examples/fishjam-chat/package.json
mironiasty: Hmm, maybe lets use `--project` then? 🤔 
(or leave as is, no strong opinion)

MiloszFilimowski: Sure!

README.md
MiloszFilimowski: It's discussed as a "NOTE" below:
> Be sure to run `npx expo prebuild --clean` and not `yarn prebuild --clean` as there's an issue with path generation for ios/.xcode.env.local

PR #113, opened 2024-09-16
[FCE-528]: Rename org.membraneframework.reactnative

MiloszFilimowski:
## Description

- Renamed android package name to `io.fishjam.reactnative`

## Motivation and Context

- We were using old package name

## How has this been tested?

- Successfully built Android.

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/android/src/main/java/com/fishjam/reactnative/AudioDeviceKind.kt
mironiasty, line 1: Should ti be `java/io/fishjam/reactnative` path?

mironiasty (approved):

PR #116, opened 2024-09-17
[FCE-539]: Fix android local deps in example project

MiloszFilimowski:
## Description

- Added example app Android plugin
- TODO: Created a follow-up issue [FCE-540](https://linear.app/swmansion/issue/FCE-540) to make both SDK and example plugins overridable, meaning we can run `expo prebuild` without `--clean` flag. 

## Motivation and Context

- Local Android SDK was not attached to example project. 

## How has this been tested?

- Successfully built Android 

## Types of changes

      not work as expected)

## Checklist:


examples/fishjam-chat/plugin/src/with-custom-config-android.ts
```diff
@@ -0,0 +1,43 @@
...
+        }
+        return match;
+      },
+    );
+
+    return config;
+  });
+};
+
+export const withCustomConfigAndroid: ConfigPlugin = (config) => {
```
mironiasty, line 38: WDYT about adding small comment in `packages/react-native-client/android/build.gradle` that this plugin is required in order for local version to be used?

examples/fishjam-chat/app.json
```diff
@@ -55,7 +55,7 @@
           }
         }
       ],
-      ["./plugin/build/development-pods-plugin.js"] // Only relevant for local development of packages/ios-client
+      ["./plugin/build/with-custom-development-paths-plugin.js"] // Only relevant for local development of packages/ios-client and packages/android-client
```
mironiasty, line 58: WDYT about changing name to better explain what it does. And move comments to plugin itself, that this is plugin required only for example app?
```suggestion
      ["./plugin/build/use-local-paths-for-native-packages.js"] 
```

examples/fishjam-chat/plugin/src/with-custom-config-ios.ts
```diff
@@ -25,8 +25,8 @@ function replaceCloudClientForMainApp(podfileContent: string) {
   return podfileContent;
 }
 
-const withCustomPodfile: ConfigPlugin = (config) => {
-  return withPodfile(config, async (config) => {
+export const withCustomConfigIOS: ConfigPlugin = (config) => {
```
mironiasty, line 28: I think we use `Ios` more often than `IOS`. But 🤷

mironiasty (approved):

mironiasty (approved):

PR #121, opened 2024-09-18
[FCE-42]: Fix simultaneously displaying 2 preview's

MiloszFilimowski:
## Description

- Add warning when someone tries to start camera again before leaving room. This API should probably be refactored in the future as it's not very intuitive. 

## Motivation and Context

- When two previews were displayed on Android, none of them worked. This was because it's not possible to initialize two camera instances on Android. 

## How has this been tested?

X.X.X)

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/utils/errorListener.ts
```diff
@@ -0,0 +1,13 @@
+import { ReceivableEvents } from '../hooks/useFishjamEvent';
+import { nativeModuleEventEmitter } from '../RNFishjamClientModule';
+
+export const initializeWarningListener = () => {
+  if (__DEV__) {
+    nativeModuleEventEmitter.addListener(
```
mironiasty, line 6: Just in case, can you wrap this code into some try/catch? Just to avoid app crashing on start just by importing our lib

PR #125, opened 2024-09-18
Disable back button and gesture only in room

maksg:
## Description

Enable back button for preview screen and disable back gesture for room screen in `fishjam-chat`.

## Motivation and Context

If I make a mistake in credentials for joining room I would like to be able to close the screen and fix it. Currently only the back gesture wasn't disabled and could be used to go back.

Additionally, I could use the gesture to go back from the room screen instead of having to tap end call button.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/examples/fishjam-chat/navigators/AppNavigator.tsx
+++ b/examples/fishjam-chat/navigators/AppNavigator.tsx
@@ -79,17 +79,21 @@ export default function AppNavigator() {
     <NavigationContainer>
       <Stack.Navigator
         initialRouteName="Home"
-        screenOptions={() => ({
-          headerBackTitleVisible: false,
-          headerBackVisible: false,
-        })}>
+        screenOptions={() => ({ headerBackTitleVisible: false })}>
         <Stack.Screen
           name="Home"
           options={{ headerShown: false }}
           component={TabNavigator}
         />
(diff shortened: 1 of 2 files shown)


MiloszFilimowski (commented):
I think we'd also need to call `leaveRoom` when this screen is closed as we need to clear any created tracks.

maksg:
I don't think that's the case as `joinRoom` is only called after tapping `Join room` button in preview screen. And going to the preview screen doesn't do much besides setting connection error to null and passing some parameters:
```
setConnectionError(null);
navigation.navigate('Preview', {
  fishjamUrl: fishjamUrl.trim(),
  participantToken: participantToken.trim(),
});
```

MiloszFilimowski:
Unfortunately, the API is not clear here. `PreviewScreen` calls `prepareCamera` which creates a local video track. This track is not removed until `leaveRoom` is called. This was the reason for this PR: https://github.com/fishjam-cloud/mobile-client-sdk/pull/121

maksg:
Ah, I see. In that case I added calling `leaveRoom` when the preview screen is supposed to close.

MiloszFilimowski (approved):

PR #139, opened 2024-09-24
[FCE-475] Call PeersUpdate event on leaveRoom

maksg:
## Description

Participants were not updating when leaving room. We should call `PeersUpdate` event after disconnecting.

### DIFF
--- a/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
+++ b/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
@@ -263,7 +263,9 @@ class RNFishjamClient(
     isScreenShareOn = false
     isConnected = false
     isCameraInitialized = false
-    fishjamClient.leave()
+    fishjamClient.leave {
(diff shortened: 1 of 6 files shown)


packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -263,7 +263,9 @@ class RNFishjamClient(
     isScreenShareOn = false
     isConnected = false
     isCameraInitialized = false
-    fishjamClient.leave()
+    fishjamClient.leave {
```
MiloszFilimowski, line 266: **question:** Why is the callback needed? Looks like this should be synchronous 🤔

maksg: On Android everything in `leave` is called in `coroutineScope.launch` so it's asynchronous. I believe that there is a reason why it was added so I didn't want to change that.
On iOS it is synchronous but I wanted it to match its Android counterpart.

PR #142, opened 2024-09-25
[FCE-593] fix displaying preview without is camera on

karkakol:
## Description

Preview displayed without making sure that `isCameraOn` resolves in blank view even if later its value changes to true

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/examples/fishjam-chat/package.json
+++ b/examples/fishjam-chat/package.json
@@ -5,8 +5,8 @@
   "scripts": {
     "prebuild": "expo prebuild",
     "start": "expo start",
-    "android": "expo run:android -d",
-    "ios": "expo run:ios -d",
+    "android": "expo run:android",
+    "ios": "expo run:ios",
     "test": "jest",
     "build": "tsc && tsc --project ./plugin",
     "lint": "sh ./scripts/lint.sh",
--- a/examples/fishjam-chat/package.json
+++ b/examples/fishjam-chat/package.json
@@ -5,8 +5,8 @@
   "scripts": {
     "prebuild": "expo prebuild",
     "start": "expo start",
-    "android": "expo run:android",
-    "ios": "expo run:ios",
+    "android": "expo run:android -d",
+    "ios": "expo run:ios -d",
     "test": "jest",
     "build": "tsc && tsc --project ./plugin",
     "lint": "sh ./scripts/lint.sh",
--- a/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
+++ b/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
@@ -12,7 +12,7 @@ class VideoPreviewView(
   RNFishjamClient.OnLocalCameraTrackChangedListener {
   private var localVideoTrack: LocalVideoTrack? = null
 
-  private fun getLocalTrackAndInitializeView() {
+  private fun trySetLocalCameraTrack() {
     localVideoTrack =
       RNFishjamClient.fishjamClient.getLocalEndpoint().tracks.values.firstOrNull { track ->
         track is LocalVideoTrack
@@ -21,7 +21,7 @@ class VideoPreviewView(
   }
 
   private fun initialize() {
-    getLocalTrackAndInitializeView()
+    trySetLocalCameraTrack()
     super.setupTrack()
   }
 
@@ -31,13 +31,13 @@ class VideoPreviewView(
   }
 
   override fun onDetachedFromWindow() {
-    RNFishjamClient.onLocalCameraTrackListeners.remove(this)
+    RNFishjamClient.localCameraTrackListeners.remove(this)
     super.onDetachedFromWindow()
     dispose()
   }
 
   override fun onAttachedToWindow() {
-    RNFishjamClient.onLocalCameraTrackListeners.add(this)
+    RNFishjamClient.localCameraTrackListeners.add(this)
     super.onAttachedToWindow()
     initialize()
   }
@@ -45,6 +45,6 @@ class VideoPreviewView(
   override fun getVideoTrack(): VideoTrack? = localVideoTrack
 
   override fun onLocalCameraTrackChanged() {
-    if (localVideoTrack == null) getLocalTrackAndInitializeView()
+    if (localVideoTrack == null) trySetLocalCameraTrack()
   }
 }
--- a/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
+++ b/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
@@ -8,16 +8,20 @@ import expo.modules.kotlin.AppContext
 class VideoPreviewView(
   context: Context,
   appContext: AppContext
-) : VideoView(context, appContext) {
+) : VideoView(context, appContext),
+  RNFishjamClient.OnLocalCameraTrackChangedListener {
   private var localVideoTrack: LocalVideoTrack? = null
 
-  private fun initialize() {
+  private fun getLocalTrackAndInitializeView() {
     localVideoTrack =
       RNFishjamClient.fishjamClient.getLocalEndpoint().tracks.values.firstOrNull { track ->
         track is LocalVideoTrack
       } as? LocalVideoTrack?
-
     videoView.let { localVideoTrack?.addRenderer(it) }
+  }
+
+  private fun initialize() {
+    getLocalTrackAndInitializeView()
     super.setupTrack()
   }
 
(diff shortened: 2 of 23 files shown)


packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
```diff
@@ -8,16 +8,20 @@ import expo.modules.kotlin.AppContext
 class VideoPreviewView(
   context: Context,
   appContext: AppContext
-) : VideoView(context, appContext) {
+) : VideoView(context, appContext),
+  RNFishjamClient.OnLocalCameraTrackChangedListener {
   private var localVideoTrack: LocalVideoTrack? = null
 
-  private fun initialize() {
+  private fun trySetLocalCameraTrack() {
```
maksg, line 15: Maybe `trySettingLocalCameraTrack` would be easier to read name?

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -13,11 +17,8 @@ class VideoPreviewView: ExpoView {
         addSubview(videoView!)
     }
 
-    override func willMove(toSuperview newSuperview: UIView?) {
-        super.willMove(toSuperview: newSuperview)
-        if newSuperview == nil {
-            localVideoTrack?.stop()
-        } else {
+    private func trySetLocalCameraTrack() {
+        DispatchQueue.main.async {
```
maksg, line 21: I think we should use `[weak self]` here. Or `[weak localVideoTrack, weak videoView]`

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -26,12 +27,28 @@ class VideoPreviewView: ExpoView {
...
-            localVideoTrack?.start()
-            videoView?.track = localVideoTrack
+            self.localVideoTrack?.start()
+            self.videoView?.track = self.localVideoTrack
+        }
+    }
+
+    override func willMove(toSuperview newSuperview: UIView?) {
+        super.willMove(toSuperview: newSuperview)
+        if newSuperview == nil {
```
maksg, line 41: Personally I like `guard let _ = newSuperview else { ... }` better because it avoids going deeper in braces but that one is up for discussion.

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -61,4 +78,10 @@ class VideoPreviewView: ExpoView {
             }
         }
     }
+
+    func onLocalCameraTrackChanged() {
+        if localVideoTrack == nil {
```
maksg, line 83: Here it could also be like this so we avoid nesting.
```
guard localVideoTrack == nil else { return }
trySettingLocalCameraTrack()
```

maksg (changes requested):

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -26,12 +27,28 @@ class VideoPreviewView: ExpoView {
...
+            self.localVideoTrack?.start()
+            self.videoView?.track = self.localVideoTrack
+        }
+    }
+
+    override func willMove(toSuperview newSuperview: UIView?) {
+        super.willMove(toSuperview: newSuperview)
+        if newSuperview == nil {
+            RNFishjamClient.localCameraTrackListeners.removeAll(where: {
+                if let view = $0 as? VideoRendererView {
```
MiloszFilimowski, line 43: **issue:** 
Shouldn't this be `VideoRendererView`? 
```suggestion
                if let view = $0 as? VideoPreviewView {
```

Also I believe you don't need to cast anything here. 
If you add `protocol OnLocalCameraTrackChangedListener: AnyObject` 

`RNFishjamClient.localCameraTrackListeners.removeAll(where: { $0 === self })` should work.

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -1,7 +1,11 @@
 import ExpoModulesCore
 import FishjamCloudClient
 
-class VideoPreviewView: ExpoView {
+internal protocol OnLocalCameraTrackChangedListener {
```
MiloszFilimowski, line 4: **nitpick:** `internal` is default so not really needed.

examples/fishjam-chat/package.json
```diff
@@ -5,8 +5,8 @@
   "scripts": {
     "prebuild": "expo prebuild",
     "start": "expo start",
-    "android": "expo run:android",
-    "ios": "expo run:ios",
+    "android": "expo run:android -d",
+    "ios": "expo run:ios -d",
```
MiloszFilimowski, line 9: **question:** What does `-d` do? And why do we need it?

MiloszFilimowski (commented):
Great work! 

I have a question regarding how we declare the listeners on Android and iOS.

Can we hide the implementation and only expose add/remove methods? Right now every other class can manipulate other listeners and it's not really the concern of the listener to handle removing itself.

Take a look at this: https://refactoring.guru/design-patterns/observer/swift/example.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/VideoPreviewView.kt
karkakol: Not sure it is easier, imho `trySetLocalCameraTrack` is more concise and aligns with common naming conventions in programming.

packages/react-native-client/ios/VideoPreviewView.swift
karkakol: I prefer to use guard when after `else` is one line of code or return statement. In this case it is not applicable.

packages/react-native-client/ios/VideoPreviewView.swift
karkakol: You are right in both cases, i will update protocol declaration.

examples/fishjam-chat/package.json
karkakol: it allows to select device to run app, including real devices.

karkakol:
> Great work!
> 
> I have a question regarding how we declare the listeners on Android and iOS.
> 
> Can we hide the implementation and only expose add/remove methods? Right now every other class can manipulate other listeners and it's not really the concern of the listener to handle removing itself.
> 
> Take a look at this: https://refactoring.guru/design-patterns/observer/swift/example.

The reason i did it is consistency. Right now on both platform we do not use properly observer design pattern. Imo it should be done globally. I will create seperate ticket for it.

karkakol:
> Great work!
> 
> I have a question regarding how we declare the listeners on Android and iOS.
> 
> Can we hide the implementation and only expose add/remove methods? Right now every other class can manipulate other listeners and it's not really the concern of the listener to handle removing itself.
> 
> Take a look at this: https://refactoring.guru/design-patterns/observer/swift/example.

Here is link to created task: https://linear.app/swmansion/issue/FCE-603/refactor-iosandroid-code-to-use-observer-pattern

MiloszFilimowski (approved):

packages/react-native-client/ios/VideoPreviewView.swift
```diff
@@ -13,11 +17,9 @@ class VideoPreviewView: ExpoView {
...
     }
 
-    override func willMove(toSuperview newSuperview: UIView?) {
-        super.willMove(toSuperview: newSuperview)
-        if newSuperview == nil {
-            localVideoTrack?.stop()
-        } else {
+    private func trySetLocalCameraTrack() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self = self else { return }
```
maksg, line 22:
```suggestion
            guard let self else { return }
```
Since Swift 5.7 we can omit the other `self`

packages/react-native-client/ios/VideoPreviewView.swift
karkakol: https://www.swiftwithvincent.com/blog/when-do-you-really-need-to-use-weak-self
From this source i assume that using [weak self] might be better, although the chances that between calling `DispatchQueue.main.async` and its execution `VideoPreviewView` will be dealocated is marginal.

maksg (approved):

examples/fishjam-chat/package.json
mironiasty: :50_50:
Can we leave it to developer to add this param? I usually prefer to have run app on default device, and use `yarn ios -d` if I want to change it.

mironiasty: :50_50:
Can we leave it to developer to add this param? I usually prefer to have run app on default device, and use yarn ios -d if I want to change it.

mironiasty (commented):
As it is not visible here:
https://github.com/fishjam-cloud/mobile-client-sdk/pull/142/files#r1775377169

examples/fishjam-chat/package.json
karkakol: i will change it

PR #155, opened 2024-09-30
[FCE-40]: Handling issues due to incorrect SDK usage in mobile app

MiloszFilimowski:
## Description

- Unified permissions flow between Android and iOS. Permissions are now requested when resource is needed upon calling `startCamera` or `startMicrophone`
- Added warning for cases when permissions are needed but not granted. 

## Motivation and Context

- Permission where not consistent between platform and not obvious in use.

## How has this been tested?

- Built and tested both Android and iOS.
- Tested screensharing and background on Android

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/ios/Utils/PermissionUtils.swift
```diff
@@ -0,0 +1,17 @@
...
+    
+    private static func requestAccessIfNeeded(for mediaType: AVMediaType) async -> Bool {
+        let status = AVCaptureDevice.authorizationStatus(for: mediaType)
+        
+        var isAuthorized = status == .authorized
+        if status == .notDetermined {
+            isAuthorized = await AVCaptureDevice.requestAccess(for: mediaType)
+        }
+        
+        return isAuthorized
```
maksg, line 15:
```suggestion
        return if status == .notDetermined {
            await AVCaptureDevice.requestAccess(for: mediaType)
        } else {
            status == .authorized
        }
```

MiloszFilimowski: Why? Status won't be updated after `requestAccess` call.

maksg (approved):

packages/react-native-client/ios/Utils/PermissionUtils.swift
```diff
@@ -0,0 +1,16 @@
+import AVFoundation
+
+class PermissionUtils {
```
karkakol, line 3: **suggestion:** 
Maybe create extension on `AVCaptureDevice` instead of seperate class?

packages/react-native-client/ios/RNFishjamClientModule.swift
```diff
@@ -86,6 +82,30 @@ public class RNFishjamClientModule: Module {
...
+            return rnFishjamClient.isCameraOn
+        }
+
+        Property("cameras") {
+            rnFishjamClient.getCaptureDevices()
+        }
+
+        Property("isScreenShareOn") {
+            return rnFishjamClient.isScreenShareOn
+        }
```
karkakol, line 107: **suggestion:** 
I know it is not your code, but if you refactored it lets be consistent and use return in every property.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -337,6 +343,11 @@ class RNFishjamClient(
   }
 
   private suspend fun startMicrophone() {
+    if (!PermissionUtils.requestMicrophonePermission(appContext = appContext)) {
```
karkakol, line 346: **suggestion:** 
appContext label can be ommited

karkakol (commented):
Good code!

I left few coments respond to them ;)

packages/react-native-client/ios/Utils/PermissionUtils.swift
MiloszFilimowski: I did it that way so that API is somewhat consistent with Android. It's easier to find corresponding code then.

packages/react-native-client/ios/RNFishjamClientModule.swift
```diff
@@ -86,6 +82,30 @@ public class RNFishjamClientModule: Module {
             return client
         }()
 
+        Name("RNFishjamClient")
+
+        Events(EmitableEvents.allEvents)
+
+        Property("peerStatus") {
+            rnFishjamClient.peerStatus.rawValue
```
karkakol: **nitpick:** 
This one also sir.

karkakol: If you are determined you can do it for also RNFishjamClientModule.kt to be consistent.

packages/react-native-client/ios/Utils/PermissionUtils.swift
karkakol: I agree.

packages/react-native-client/ios/RNFishjamClientModule.swift
MiloszFilimowski: Okay!

karkakol (approved):
image

PR #163, opened 2024-10-04
[FCE-641]: Android screensharing fix

MiloszFilimowski:
## Description

- Fixed screensharing to wait for the service to become active before launching screen cast
- Refactor RN code to make it easier for users to use. Right now you only have to call `toogleScreenCast` and use `useForegroundService` hook to make it work. No permissions dance required. 

## Motivation and Context

- Foreground service was crashing the app

## How has this been tested?

- Tested on Android API level 35

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/hooks/useScreenShare.ts
```diff
@@ -108,6 +91,29 @@ export function useScreenShare() {
     return 'denied';
   }, []);
 
+  const toggleScreenShare = useCallback(
+    async (screenShareOptions: Partial<ScreenShareOptions> = {}) => {
+      if (Platform.OS === 'android' && !isScreenShareOn) {
+        (await handleScreenSharePermission()) == 'granted' &&
```
graszka22, line 97: what if user doesn't grant permission?

MiloszFilimowski: A warning will be emitted to the developer: "No permission to start screen share, call handleScreenSharePermission first."

graszka22: Shouldn't developer be able to handle this gracefully, like displaying a modal to the user "hey, you didn't grant the permission to screen share, go to settings if you change your mind" or something like that?

MiloszFilimowski: `toggleScreenShare` is a promise that the developer can catch. However, I can see now this this could be handled better with an error code rather then a message.

MiloszFilimowski: Ok, I checked that this is a class, so the exception code is `ERR_MISSING_SCREEN_SHARE_PERMISSION` so everything should be fine.

graszka22 (approved):

packages/react-native-client/src/index.tsx
```diff
@@ -51,15 +51,15 @@ export { VideoPreviewView } from './components/VideoPreviewView';
...
-  stopForegroundService,
-} from './utils/foregroundService';
-
 export type { PeerStatus } from './hooks/usePeerStatus';
 export { usePeerStatus } from './hooks/usePeerStatus';
 
+export type {
+  ForegroundServiceNotificationConfig,
+  ForegroundServicePermissionsConfig,
+} from './types';
```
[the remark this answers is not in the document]
MiloszFilimowski: Sure!

packages/react-native-client/src/hooks/useForegroundService.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const useForegroundService = ({
+  enableCamera,
+  enableMicrophone,
+  ...restOptions
+}: ForegroundServiceNotificationConfig &
+  ForegroundServicePermissionsConfig) => {
+  /* eslint-disable react-hooks/rules-of-hooks */
+  if (Platform.OS !== 'android') {
```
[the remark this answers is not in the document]
MiloszFilimowski: I think you submitted your comments after my changes. I refactor this `if` into an export with `Platform.select`. Personally, I'm not a fan of `android.ts` and `ios.ts`.

examples/fishjam-chat/screens/RoomScreen.tsx
```diff
@@ -47,40 +47,27 @@ const RoomScreen = ({ navigation, route }: Props) => {
...
-        return;
-      }
-      if ((await handleScreenSharePermission()) != 'granted') {
-        return;
-      }
-    },
-    [handleScreenSharePermission],
-  );
-
   const onToggleScreenShare = useCallback(async () => {
```
karkakol, line 66: **nitpick:** 
Name of the function suggest that it is called when screenShare is toggled, meanwhile it is calling toggleScreenShare. I think name of this function should be changed to  `toggleScreenShare`.

packages/react-native-client/src/hooks/useScreenShare.ts
```diff
@@ -108,6 +91,29 @@ export function useScreenShare() {
...
+      const options = {
+        ...screenShareOptions,
+        screenShareMetadata: {
+          displayName: 'presenting',
+          type: 'screensharing' as const,
+          active: !isScreenShareOn,
+        },
+      };
+      await RNFishjamClientModule.toggleScreenShare(options);
+      screenShareSimulcastConfig = defaultSimulcastConfig(); //to do: sync with camera settings
```
karkakol, line 111: **suggestion:** 
What that comment is about?

MiloszFilimowski: I have no idea, it shows as changed because of refactoring. 

Maybe @mironiasty remembers? 
https://github.com/fishjam-cloud/mobile-client-sdk/pull/84/files

examples/fishjam-chat/screens/RoomScreen.tsx
karkakol: Let it be. I see that this code is not even your so you can leave it as it is.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/Errors.kt
```diff
@@ -11,8 +11,7 @@ class ConnectionError(
   reason: AuthError
 ) : CodedException(message = "Connection error: ${reason.error}")
 
-class MissingScreenSharePermission :
-  CodedException(message = "No permission to start screen share, call handleScreenSharePermission first.")
+class MissingScreenSharePermission : CodedException(message = "No permission to start screen share.")
```
karkakol, line 14: **nitpick:** 
Lets message be omitted in constructor.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -74,6 +75,8 @@ class RNFishjamClient(
       emitEvent(event, mapOf(event.name to value))
     }
 
+  private var foregroundServiceManager: ForegroundServiceManager? = null
```
karkakol, line 78: **question:** 
1. Can foregrund service be called when user is not connected to rom?
2. Does `foregroundServiceManager` must be nullable? From its curent use i think it can be not nullable

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/foreground/ForegroundServiceManager.kt
```diff
@@ -0,0 +1,147 @@
+package io.fishjam.reactnative.foreground
```
karkakol, line 1: **suggestion:** 
Package name is a little bit not clear. Maybe sth like foregroundService.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/utils/PermissionUtils.kt
karkakol, line 1: **suggestion:** 
I think all methods can be static.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/Errors.kt
MiloszFilimowski: Why?

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
MiloszFilimowski: 1. Yes
2. How would that work? `configureForegroundService` must be called to instantiate a `ForegroundServiceManager` instance.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/utils/PermissionUtils.kt
MiloszFilimowski: They are all static. If you look at the declaration it's an `object` not a class.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/Errors.kt
karkakol: It's shorter :P

MiloszFilimowski: I like explicit parameters, especially when it's not obvious what it is. It's also consistent with how other errors are declared.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/utils/PermissionUtils.kt
karkakol: U right! 👍

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/Errors.kt
karkakol: From kotlin coding conventions:
```
Use the named argument syntax when a method takes multiple parameters of the same primitive type, or for parameters of Boolean type, unless the meaning of all parameters is absolutely clear from context.
```

`CodedException` takes two arguments which are trivial for exceptions : `message` and `cause`. Imo there is no point in redundant repeating parameter name.

karkakol: I see that consistency is more important, so it can be changed in future during some android refactoring.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
karkakol: Some default can be passed to initialize it in constructor or in property. Then when `configureForegroundService` is called manager would be updated.

karkakol: But this is my point of view. If sth can be not null I would prefer do it that way.

MiloszFilimowski: It doesn't make sense. This is the reason there is a configure method. It will just move the null from `RNFishjamClient` to `ForegroundServiceManager` and we explicitly want to have configuration set once. It also provides additional info if the manager is already configured.

karkakol: It does not move null to `ForegraoundServiceManager`. Consider this code:
```suggestion
  class ForegroundServiceNotificationConfig : Record {
  @Field
  val channelId: String = "default_value"

  @Field
  val channelName: String = "default_value"
  @Field
  val notificationContent: String = "default_value"

  @Field
  val notificationTitle: String = "default_value"
}

  private var foregroundServiceManager: ForegroundServiceManager = ForegroundServiceManager(appContext!!, ForegroundServiceNotificationConfig()) 
```

In this case there is no null values, it could be directly used by developer without knowing what are specified properties in `ForegroundServiceNotificationConfig`, and only if user want to change it he would do it. Also `configureForegroundService` does not have to be called by user.

MiloszFilimowski: > Also configureForegroundService does not have to be called by user
The user never calls it, it's our internal method that should be called once the hook is initialized. 

In my opinion, you're trying to fit a use case, to justify having a non-null value even when it makes perfect sense to have one. With your approach:
1. I'd have to create another method in the manager to reconfigure the config 
2. Create another variable to check if it was configured (it should not be changed once configured).

karkakol: 1. I think that you can just create new manager (correct me if i am wrong)
2. Actually i think that notification text should be reconfigurable. User should be able to change notification texts like e.g. messenger does when you call somebody. If i understand how it works now it is not possible to change it when app is running.

MiloszFilimowski: 1. You can't because foreground service is bound to the manager and it keeps track of it.
2. This would require more work and can be done as a separate task

karkakol: Create task for it, please. From me green light to merging!

karkakol (approved):
👍

PR #164, opened 2024-10-04
Karkakol/add app screen sharing

karkakol:
## Description

Describe your changes in detail

## Motivation and Context

Why is this change required? What problem does it solve? If it fixes an open
issue, please link to the issue here.

## How has this been tested?

X.X.X)

## Types of changes

      not work as expected)

## Checklist:


packages/ios-client/Sources/FishjamClient/media/Tracks/LocalAppScreenShareTrack.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    ) {
+        mediaTrack.shouldReceive = false
+        self.videoParameters = videoParameters
+        self.capturer = capturer
+        self.videoSource = videoSource
+        super.init(mediaTrack: mediaTrack, endpointId: endpointId, rtcEngineId: nil, metadata: metadata)
+    }
+
+    convenience internal init(mediaTrack: RTCVideoTrack, oldTrack: LocalAppScreenShareTrack) {
```
MiloszFilimowski, line 21: **nitpick:** all `internal` declarations are redundant.

packages/react-native-client/ios/RNFishjamClient.swift
```diff
@@ -327,6 +344,10 @@ class RNFishjamClient: FishjamClientListener {
     func toggleScreenShare(screenShareOptions: ScreenShareOptions) throws {
         try ensureCreated()
         try ensureConnected()
+        guard isAppScreenShareOn == false else {
```
MiloszFilimowski, line 347: **suggestion:** 
```suggestion
        guard !isAppScreenShareOn else {
```

packages/react-native-client/src/hooks/useAppScreenShare.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @group Hooks
+ */
+export function useAppScreenShare(): {
+  isAppScreenShareOn: boolean;
+  simulcastConfig: SimulcastConfig;
+  toggleAppScreenShare: (
+    screenShareOptions?: Partial<ScreenShareOptions>,
+  ) => Promise<void>;
+} | null {
+  if (Platform.OS === 'android') return null;
```
MiloszFilimowski, line 28: **issue:** 
```suggestion
  if (Platform.OS !== 'ios') return null;
```

What do you think about returning default values here instead of null? This way we would have a more pleasant API for the user, without having to unwrap nulls.

packages/react-native-client/ios/RNFishjamClient.swift
karkakol: In my opinion it is more readable to use in this case comparison to false than negating the value.

examples/fishjam-chat/screens/RoomScreen/RoomScreen.tsx
```diff
@@ -143,19 +141,7 @@ const RoomScreen = ({ navigation, route }: Props) => {
...
-                ? 'cellphone-screenshot'
-                : 'cellphone-off'
-            }
-            onPress={() => {
-              iosAppScreenShare?.toggleAppScreenShare();
-            }}
-          />
-        ) : null}
-
+        {Platform.OS === 'ios' && <ToggleAppScreenButton />}
```
MiloszFilimowski, line 144: **nitpick:** I think it would be better if `ToggleAppScreenButton` handled this case. It's not possible to use it outside of iOS so I think it should return null for other platforms.

packages/ios-client/Sources/FishjamClient/media/Tracks/LocalAppScreenShareTrack.swift
```diff
@@ -18,7 +18,7 @@ public class LocalAppScreenShareTrack: VideoTrack, LocalTrack {
         super.init(mediaTrack: mediaTrack, endpointId: endpointId, rtcEngineId: nil, metadata: metadata)
     }
 
-    convenience internal init(mediaTrack: RTCVideoTrack, oldTrack: LocalAppScreenShareTrack) {
+    convenience init(mediaTrack: RTCVideoTrack, oldTrack: LocalAppScreenShareTrack) {
```
MiloszFilimowski, line 21: **chore:** Can you please also remove all other internals? 🙏

examples/fishjam-chat/screens/RoomScreen/ToggleAppScreenShare.ios.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useCallback } from 'react';
+import { roomScreenLabels } from '../../types/ComponentLabels';
+import { InCallButton } from '../../components';
+import { useAppScreenShare } from '@fishjam-cloud/react-native-client';
+
+export const ToggleAppScreenButton = () => {
+  const iosAppScreenShare = useAppScreenShare();
```
MiloszFilimowski, line 7: **suggestion:** If it's not optional any more, I guess you can use `{ toggleAppScreenShare }`

examples/fishjam-chat/screens/RoomScreen/RoomScreen.tsx
karkakol: I do not understand your point. `ToggleAppScreenButton` handles it, if some one use it he would get button which does nothing. Lets remember it is our example app.
Also from layout view it is more important to parent view to decide if child should be displayed than child decides for itself.

packages/ios-client/Sources/FishjamClient/media/Tracks/LocalAppScreenShareTrack.swift
karkakol: Lets do it in seperate ticket when swift code would be refactored.

MiloszFilimowski (approved):

PR #170, opened 2024-10-09
[FCE-566] Add prop to disable video player when not visible to user

maksg:
## Description

This PR adds prop which can disable video player if out of screen.

## Motivation and Context

If the user can't see the video player then there's no need to waste device's power to receive the track.

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/components/VideoRendererView.tsx
```diff
@@ -14,6 +14,7 @@ export type VideoRendererProps = {
    * @default `FILL`
    */
   videoLayout?: VideoLayout;
+  alwaysRender?: boolean;
```
MiloszFilimowski, line 17: **suggestion:** Can you please document and describe what this property does, when and why should user set it.

MiloszFilimowski (approved):
Great work! Smart solution with a timer! 
Just one comment about docs.

MiloszFilimowski:
I'm wondering if we could further reduce energy usage if we used just a single timer for all of the views. AFAIK timers are quite expensive: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/MinimizeTimerUse.html

examples/fishjam-chat/components/VideosGrid.tsx
```diff
@@ -55,6 +55,7 @@ export default function VideosGrid({ tracks }: Props) {
           <VideoRendererView
             trackId={track.id}
             videoLayout="FIT"
+            alwaysRender={false}
```
karkakol, line 58: **suggestion:** 
I think that this name can be confusing and could be renamed to something self explaining`shouldRenderIfViewIsNotAtScreen`, `renderOffscreen` or anything you find not confiusing.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/media/VideoTrack.kt
```diff
@@ -24,4 +24,8 @@ open class VideoTrack(
   fun removeRenderer(renderer: VideoSink) {
     this.videoTrack.removeSink(renderer)
   }
+
+  fun setShouldReceive(shouldReceive: Boolean) {
```
karkakol, line 28: **suggestion:** 
I think that this function could be named shorter `shouldReceive`.

karkakol (commented):
I like your code and your solution!
I was wondering if our package should be be responsible for handling track receiving, and maybe this responsibility should be done by app developer, and our api should only provide a way to decide if track is received or not. I do not have strong opinion about that. Let me know what do you think.
Also I was wondering if this solution works properly when used with navigators like stack etc. Have you tested it?

maksg:
> I'm wondering if we could further reduce energy usage if we used just a single timer for all of the views. AFAIK timers are quite expensive: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/EnergyGuide-iOS/MinimizeTimerUse.html

Unfortunately it might not be possible without some major refactor. `VideoRendererView` is created directly from React Native and because of that all created video players have no way of communicating with each other.

However the performance gain from disabling video players is greater than the penalty for using so many timers and the timers don't run with `shouldRenderIfNotVisible` set to `true` so the performance shouldn't be worse in any case.

MiloszFilimowski:
>Unfortunately it might not be possible without some major refactor. VideoRendererView is created directly from React Native and because of that all created video players have no way of communicating with each other.

What if we create a singleton instance of something like `TimerManager/SharedVideoTimer` that `VideoRenderView`s could subscribe to. When the subscribe we increase the counter, when they unsubscribe (or deallocate) and the count goes to zero, timer stops. We could notify them with some kind of a listener. What do you think?

maksg:
@MiloszFilimowski @karkakol I've resolved the crash from Android e2e tests and there is a lot of changes so please take a look again.

It seems that iOS app never deallocated `VideoRendererView` and because of that timer couldn't be removed as well. I created a workaround for that but it may be a React Native thing that it's still in memory.

Also sometimes there is a crash related to screen capture on Android but it's random and doesn't seem to be related to my changes. It crashes on `capturer.stopCapture()` and says that it has already been disposed. This is something to be investigated in the future.

karkakol:
> @MiloszFilimowski @karkakol I've resolved the crash from Android e2e tests and there is a lot of changes so please take a look again.
> 
> It seems that iOS app never deallocated `VideoRendererView` and because of that timer couldn't be removed as well. I created a workaround for that but it may be a React Native thing that it's still in memory.
> 
> Also sometimes there is a crash related to screen capture on Android but it's random and doesn't seem to be related to my changes. It crashes on `capturer.stopCapture()` and says that it has already been disposed. This is something to be investigated in the future.

Are you sure that error on android is reproducable on `main`? If so answer me and i will approve.

packages/react-native-client/src/components/VideoRendererView.tsx
```diff
@@ -14,6 +14,10 @@ export type VideoRendererProps = {
    * @default `FILL`
    */
   videoLayout?: VideoLayout;
+  /**
+   * Whether the video should be rendered if it's off screen or hidden.
+   */
+  shouldRenderIfNotVisible?: boolean;
```
mironiasty, line 20: I would consider changing name of this props, to something more like:
`skipRenderOutsideVisibleArea`
This way:
- there is no negation in prop name (it is often confusing)
- props will be opt in (as you will have to set it to true, for this functionality to be used)

mironiasty (commented):
What happens with audio track related to video that is not rendered? Is it still played?

maksg:
> What happens with audio track related to video that is not rendered? Is it still played?

@mironiasty Yep, it's still played. Just checked that

mironiasty (approved):
LGTM

PR #171, opened 2024-10-09
[FCE-666] Add videoroom support in mobile app

mironiasty:
## Description

Update example app with new Tab that would allow to connect with VideoRoom.

## Motivation and Context

It will be now possible to create build that can connect directly with VideoRoom (without need to know URLs) and allow to test mobile SDK during meetings.

## How has this been tested?

App was build and run on real device and tested with VideoRoom.
Note that during testing, bug with audio was found. See FCE-689

## Types of changes

      not work as expected)

## Checklist:

## Screenshots (if appropriate)
Screenshot_20241009_173957_FishjamChat


examples/fishjam-chat/app.json
```diff
@@ -38,7 +38,8 @@
       ]
     },
     "androidStatusBar": {
-      "hidden": true
+      "backgroundColor": "#BFE7F8",
```
karkakol, line 41: **suggestion:** 
use `BrandColors.seaBlue40` from `Colors.ts`

karkakol (approved):
image

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const onTapConnectButton = async () => {
+    try {
+      setConnectionError(null);
+      setLoading(true);
+      const roomManagerUrl =
+        env === 'staging'
+          ? process.env.EXPO_PUBLIC_VIDEOROOM_STAGING_ROOM_MANAGER!
+          : process.env.EXPO_PUBLIC_VIDEOROOM_PRODUCTION_ROOM_MANAGER!;
+      saveStorageData({ env, roomName, userName });
```
MiloszFilimowski, line 89: **question:** What is the purpose of the example app saving those details? Isn't it supposed to be as simple as possible?

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+      );
+
+      navigation.navigate('Preview', {
+        userName,
+        fishjamUrl,
+        peerToken: token,
+      });
+    } catch (e) {
+      const message =
+        'message' in (e as Error) ? (e as Error).message : 'Unknown error';
```
MiloszFilimowski, line 104: **suggestion:** Why not just `e.message ?? 'Unknown error'` Those type casts don't provide any additional value in this case, IMO.

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+    !!process.env.EXPO_PUBLIC_VIDEOROOM_PRODUCTION_ROOM_MANAGER
+  );
+}
+
+/**
+ * Connect with the VideoRoom - our example service for video conferences
+ */
+export default function ConnectScreen({ navigation }: Props) {
+  const [connectionError, setConnectionError] = useState<string | null>(null);
+  const [env, setEnv] = useState<VideoRoomEnv>('staging');
```
MiloszFilimowski, line 60: **nitpick:** Isn't env a bit too ambiguous? Env of what?

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+    gap: 24,
+  },
+  errorMessage: {
+    color: 'black',
+    textAlign: 'center',
+    margin: 25,
+    fontSize: 20,
+  },
+  logo: {
+    width: windowWidth - 40,
```
MiloszFilimowski, line 184: **suggestion:** Why not just inline `Dimensions.get('window').width` here? It's not reused anywhere.

examples/fishjam-chat/app.json
mironiasty: That will be difficult, as this is JSON file - and there is no easy way to import consts here :(

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
mironiasty: `e` is `unknown` type, so you have to cast first to ready `message` property.

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
mironiasty: It was just copy from other screen. But indeed - let's update it.

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
mironiasty: I also have problem with this. I would like to keep this app as simple as possible. But also for our internal use, it would be useful to have some app to test VIdeoRoom. So either we can create separate app for that. Or we can reuse example. And if someone would like to use this app for real VideoRoom - storing data will be super useful.

Another idea I had, was to move files around in project, to hide code related to VideoRoom (but to still keep it here). I'm open for any suggestions

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
MiloszFilimowski: I don't have a strong option about it, but in that case, you could declare e as `any`. We don't really gain anything here from this typecast. 

If you would like to stick with typecast, I suggest a shorter version which seems more concise to me:
`const message = e instanceof Error ? e.message : 'Unknown error';` 

Both versions have, however, the disadvantage of passing empty/undefined messages only if the property is specified.

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
MiloszFilimowski: The only thing I don't understand is why we need to store `roomName` and `userName`. Can't this be hard coded when we use it? Like we do with other screens?

MiloszFilimowski (approved):

maksg (approved):

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
mironiasty: What values would like to hardcode here? It is specific to user, I guess (your name and recently used room)

examples/fishjam-chat/screens/ConnectWithVideoRoomScreen.tsx
mironiasty: :50_50:
My personal experience is that having `catch(error: any)` by default (`useUnknownInCatchVariables off`) could create additional problems. So as might create more code, in long term it is helpful.
And this is example app, so let's show good practices where we can ;)

PR #173, opened 2024-10-09
[FCE-667] Remove protocol conformances resulting in warnings

maksg:
## Description

Remove protocol conformances resulting in `extension declares a conformance of imported type '[...]' to imported protocol '[...]'; this will not behave correctly if the owners of '[...]' introduce this conformance in the future` warnings

## Motivation and Context

We should avoid adding such protocol conformances since it may result in issues in the future. ~~And because because `Dimensions` is only compared once then it seems better to just do the comparison in place.~~
`Dimensions` would work better as a wrapper of `CMVideoDimensions`

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
+++ b/packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
@@ -1,5 +1,13 @@
 import Promises
 
+class CommandQueueError: Error {
+    let message: String
+
+    init(message: String) {
+        self.message = message
+    }
+}
+
 internal class CommandsQueue {
     var clientState: ClientState = ClientState.CREATED
     private var commandsQueue: [Command] = []
@@ -40,7 +48,7 @@ internal class CommandsQueue {
     func clear() {
         clientState = .CREATED
         commandsQueue.forEach { command in
-            command.promise.reject("Command queue was cleared")
+            command.promise.reject(CommandQueueError(message: "Command queue was cleared"))
         }
         commandsQueue.removeAll()
     }
--- a/packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
+++ b/packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
@@ -3,7 +3,7 @@ import Promises
 class CommandQueueError: Error {
     let message: String
 
-    init(message: String) {
+    init(_ message: String) {
         self.message = message
     }
 }
@@ -48,7 +48,7 @@ internal class CommandsQueue {
     func clear() {
         clientState = .CREATED
         commandsQueue.forEach { command in
-            command.promise.reject(CommandQueueError(message: "Command queue was cleared"))
+            command.promise.reject(CommandQueueError("Command queue was cleared"))
         }
         commandsQueue.removeAll()
     }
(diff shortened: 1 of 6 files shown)


mironiasty (approved):

packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
```diff
@@ -1,5 +1,13 @@
 import Promises
 
+class CommandQueueError: Error {
+    let message: String
+
+    init(message: String) {
```
karkakol, line 6: **suggestion:** 
To keep it shorter i suggest:
```suggestion
    init(_ message: String) {
```

packages/ios-client/Sources/FishjamClient/media/Dimensions.swift
```diff
@@ -9,12 +9,6 @@ extension Dimensions {
     public static let aspect4By3 = 4.0 / 3.0
 }
 
-extension Dimensions: Equatable {
-    public static func == (lhs: Dimensions, rhs: Dimensions) -> Bool {
```
karkakol, line 13: **suggestion:** 
Maybe instead od overloading equal operator write method named `equal` ?

karkakol: Then it could be easier to use in future

maksg: I thought about it but I decided against it. While it's true that it might be used in the future, this future might as well never come so why keep additional code just in case. Also naming function `equal` might be confusing if there could be an operator function.

However now that I think of it, it will be better if I just wrap this type in our own type instead of typealiasing

packages/ios-client/Sources/FishjamClient/CommandsQueue.swift
maksg: I'm not too sure about that one. It will be shorter, true but it may not read that good that it's supposed to be a message

packages/ios-client/Sources/FishjamClient/media/Dimensions.swift
```diff
@@ -1,23 +1,31 @@
...
+
+    public init(width: Int32, height: Int32) {
+        self.dimensions = CMVideoDimensions(width: width, height: height)
+    }
+
+    public init(_ dimensions: CMVideoDimensions) {
+        self.dimensions = dimensions
+    }
+
+    /// Swaps height with width.
```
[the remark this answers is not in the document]
maksg: Unfortunately I can't modify the `dimensions` properties because it's immutable. I left it as it was because earlier it also returned new struct. But I'll think about a better wording for that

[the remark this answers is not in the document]
karkakol: Sth like `returns struct with flipped dimensions` should be enough.  Or you can use mutating function to do swap.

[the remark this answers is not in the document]
maksg: I've already commited updated wording but thanks for the suggestion!

karkakol (approved):
taco_hulk_approves

PR #174, opened 2024-10-09
[FCE-30] fix vad

karkakol:
## Description

Fix vad and utilize it in fishjam-chat. VAD status is taken from fishjam, but i discovered it is possible to implement it on native side by working with RTCAudioTrack. Currently when single VAD status changes all endpoints are emitted. This solution might be expensive but it is very simple and it works. In future it might be a good idea to just emit single changed track.

## Motivation and Context

VAD is cool

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/examples/fishjam-chat/components/VideosGrid.tsx
+++ b/examples/fishjam-chat/components/VideosGrid.tsx
@@ -2,44 +2,64 @@ import {
   Track,
   VideoRendererView,
   Peer,
+  TrackType,
 } from '@fishjam-cloud/react-native-client';
-import React from 'react';
+import React, { useCallback } from 'react';
 import { FlatList, StyleSheet, View } from 'react-native';
 
 import { roomScreenLabels } from '../types/ComponentLabels';
 import { BrandColors } from '../utils/Colors';
 import Typo from './Typo';
 import { PeerMetadata } from '../types/metadata';
+import VADIcon from './VADIcon';
+import { AudioTrack } from '@fishjam-cloud/react-native-client/build/hooks/usePeers';
 
 type Props = {
-  tracks: GridTrack[];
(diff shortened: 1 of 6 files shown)


packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
```diff
@@ -750,7 +750,7 @@ extension FishjamClientInternal: RTCEngineListener {
     }
 
     func onVadNotification(trackId: String, status: String) {
-        guard let track = getTrack(trackId: trackId) else {
+        guard let track = getTrackWithRtcEngineId(trackId: trackId) else {
```
maksg, line 753: Maybe we should put `as? RemoteAudioTrack` here instead of casting it twice later on. Seems like this is how it works on Android as well

examples/fishjam-chat/components/VideosGrid.tsx
```diff
@@ -2,44 +2,64 @@ import {
...
-import React from 'react';
+import React, { useCallback } from 'react';
 import { FlatList, StyleSheet, View } from 'react-native';
 
 import { roomScreenLabels } from '../types/ComponentLabels';
 import { BrandColors } from '../utils/Colors';
 import Typo from './Typo';
 import { PeerMetadata } from '../types/metadata';
+import VADIcon from './VADIcon';
+import { AudioTrack } from '@fishjam-cloud/react-native-client/build/hooks/usePeers';
```
MiloszFilimowski, line 15: **issue:** This import looks weird. I think it should be a part of `@fishjam-cloud/react-native-client`

maksg (approved):

MiloszFilimowski (approved):

PR #176, opened 2024-10-10
[FCE-621] Add jitpack cache warmup

mironiasty:
## Description

When we publish new release (tag), we have to push libraries to CocoaPods and npm. But jitpack works another way, and it pulls our code from repo.

This PR is adding step to let know Jitpack that new version was released.

This PR also adds formatting for .github files.

## Motivation and Context

Right now, when we publish new version, first user that will try to build app will get error - as Android won't work. There are few minutes required after first request for JitPack to cache everything. With this change, it should work right away.

## How has this been tested?

To fully test this addition, new version has to be released. But I should not break any existing flow.

## Types of changes

      not work as expected)

## Checklist:


.github/workflows/release.yml
```diff
@@ -26,11 +26,18 @@ jobs:
...
+          registry-url: "https://registry.npmjs.org"
       - run: yarn
       - run: yarn prepare
-      - run: yarn npm publish --access public 
+      - run: yarn npm publish --access public
         working-directory: packages/react-native-client
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  publish-android-client:
+    # actually it just warmup jitpack cache, so android client is available sooner
```
karkakol, line 38: **question:** 
What does this comment mean?

karkakol (commented):
If i understand properly it tries to fetch released version from jitpack and this triggers build so it is available faster?

.github/workflows/release.yml
mironiasty: >If i understand properly it tries to fetch released version from jitpack and this triggers build so it is available faster?

That is correct
 
>What does this comment mean?

This means that this is not `publish` action, but just ask jitpack to build/cache package. I will think on how to phrase it better

MiloszFilimowski (approved):

.github/workflows/release.yml
karkakol: I think that sth like this:
```
Packages built by Jitpack are not immediately available upon the creation of a tag or branch. Instead, availability occurs when an initial attempt is made to download the specified tag or branch. This mechanism ensures that Jitpack only builds a package when it is first requested. This approach efficiently handles the generation of newly released versions.
```

mironiasty: I've updated comment

MiloszFilimowski (approved):

karkakol (approved):
taco_hulk_approves

PR #177, opened 2024-10-10
[FCE-555] Add ex-webrtc support

graszka22:
## Description

This PR adds support for ex-webrtc backend. 
On both platforms:
- ~removed `restartIce()` call - it causes issues with connectivity~ => `restartIce` is needed for non exwebrtc.
- changes ice transport type from `.relay` to .`all`
- conforms to protocol when sending candidates
- disables simulcast in example app - it's currently not supported, but maybe it will be supported in the future, thus I'm not changing the api, it's disabled by default anyway

NEW changes:
- Removed `endpointType` as it was unused and only created confusion about what should be set. 
- ExWebRTC/WebRTC is now decided based upon if turn servers are passed (same as it's done on Web)

On Android:
- fixes a (backend) crash when client sends candidate before sending sdp offer by queueing the candidates - on iOS it's not needed as sdp offer is created on the same thread as candidates. In my opinion it's more of a backend issue but it's quicker to workaround it on our side.
- fixes a bug when using `changeWebRTCLoggingSeverity` (found out when debugging something)
- changes number strategy when serializing to json, so that it sends integers instead of floats

On iOS:
- removed some dead code with `queuedRemoteCandidates`

## Motivation and Context

https://linear.app/swmansion/issue/FCE-555/investigate-work-required-to-support-elixirwebrtc-in-mobile-app

## How has this been tested?

- Get a backend with ex-webrtc (built locally from `ex_webrtc_experimental` of fishjam or deployed to fishjam.work)
- Open fishjam dashboard
- Test only on physical devices, emulators won't work - currently backend has fixed h264 encoding which is unsupported on android emulator
- Test different combinations: two mobiles, mobile + web
- Test connecting in different order, screencast, disconnecting and reconnecting, adding and removing tracks


packages/ios-client/Sources/FishjamClient/events/Event.swift
```diff
@@ -319,7 +319,6 @@ struct ConnectedEvent: ReceivableEvent, Codable {
 struct EndpointAddedEvent: ReceivableEvent, Codable {
     struct Data: Codable {
         let id: String
-        let type: EndpointType
```
graszka22, line 318: On iOS if server sends data with `type` and it isn't included here will it fail to decode and crash? As far as I remeber on iOS json decoder is strict about such things?

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -202,7 +202,7 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
 
     /// Parses a list of turn servers and sets them up as `iceServers` that can be used for `RTCPeerConnection` ceration.
     private func setTurnServers(_ turnServers: [OfferDataEvent.TurnServer]) {
-        config?.iceTransportPolicy = .relay
+        let isExWebrtc = turnServers.count == 0
```
graszka22: nice trick 🤔

graszka22:
one question and otherwise LGTM 👍

packages/ios-client/Sources/FishjamClient/events/Event.swift
MiloszFilimowski: It won't fail (I tested it). 
The default key decoding strategy ignores keys that are not present. 

It will fail if the server removes a key we require, but there is a catch block that will result in a message printing.

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -238,14 +233,16 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
...
     }
 
     // Default ICE server when no turn servers are specified
     private static func defaultIceServer() -> RTCIceServer {
-        let iceUrl = "stun:stun.l.google.com:19302"
+        let iceUrl1 = "stun:stun.l.google.com:19302"
+        let iceUrl2 = "stun:stun.l.google.com:5349"
 
-        return RTCIceServer(urlStrings: [iceUrl])
+        return RTCIceServer(urlStrings: [iceUrl1, iceUrl2])
```
mironiasty, line 245: WDYT: 
```suggestion
        let iceUrls = [
            "stun:stun.l.google.com:19302",
            "stun:stun.l.google.com:5349",
        ]

        return RTCIceServer(urlStrings: iceUrls)
```

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -201,15 +202,9 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
...
-    private func drainCandidates() {
-        for candidate in queuedRemoteCandidates {
-            connection?.add(candidate, completionHandler: { _ in })
-        }
-    }
-
     /// Parses a list of turn servers and sets them up as `iceServers` that can be used for `RTCPeerConnection` ceration.
     private func setTurnServers(_ turnServers: [OfferDataEvent.TurnServer]) {
-        config?.iceTransportPolicy = .relay
+        isExWebrtc = turnServers.count == 0
```
mironiasty, line 207: Does it need to be class level variable? It looks like it can be defined locally in `getSdpOffer`, as it used only there 🤔

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
```diff
@@ -372,17 +393,22 @@ internal class PeerConnectionManager(
...
     if (peerConnection == null) {
       setupPeerConnection(localTracks)
       needsRestart = false
     }
+
     peerConnectionMutex.withLock {
       val pc = peerConnection!!
 
-      if (needsRestart) {
+      if (needsRestart && !isExWebrtc) {
```
mironiasty, line 411: Maybe we can 'calculate' this value inside this function too, instead of keeping it in class context? 🤔

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
```diff
@@ -263,28 +271,41 @@ internal class PeerConnectionManager(
...
+          val url =
+            listOf(
+              "turn",
+              ":",
+              it.serverAddr,
+              ":",
+              it.serverPort.toString(),
+              "?transport=",
+              it.transport
+            ).joinToString("")
```
mironiasty, line 291: Can string template be used here? 🤔

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
MiloszFilimowski: Good point

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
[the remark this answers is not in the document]
MiloszFilimowski: Please refresh the page, I already use `isEmpty` :D

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -403,8 +398,9 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
         guard let pc = connection else {
             return
         }
-
-        if needsRestart {
+        
+
```
maksg, line 402: 2 empty lines

mironiasty (approved):
:LGTM:

PR #192, opened 2024-10-17
[FCE-729] Update GH Actions

mironiasty:
## Description

Update GH actions to be more up to date with current state of repo
Update how iOS linting scripts work

## Motivation and Context

To keep our scripts in better state (and fix small issues found during test release)

## How has this been tested?

I've run scripts locally. Also They are run together with this PR.

## Types of changes

      not work as expected)

## Checklist:


.github/workflows/ios_lint_test.yaml
mironiasty, line 1: This GH action is now duplicating content of existing one that lints everything

.github/workflows/react_native_publish_docs.yaml
mironiasty, line 1: This action no longer works (nor is needed - we do not have iOS docs)

.github/workflows/release.yml
```diff
@@ -35,6 +36,7 @@ jobs:
     # This mechanism ensures that Jitpack only builds a package when it is first requested and efficiently handles the generation of newly released versions.
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v4
```
mironiasty, line 39: I think this part was missing - that is why it was not working with 0.4.0 release.
As a side note: I added jitpack webhooks to repo, maybe they will work too?

packages/ios-client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@fishjam-cloud/ios-client",
   "scripts": {
-    "lint": "swift-format format -i -r ./Sources/ --configuration swift-format-config.json",
-    "lint:check": "swift-format lint -r -s ./Sources/ --configuration swift-format-config.json"
+    "lint": "sh ./scripts/lint.sh",
+    "lint:check": "sh ./scripts/lint_check.sh"
```
mironiasty, line 5: I think that for iOS both `lint` and `lint:check` should do the same thing. There is now (easy) way to do dry-run for files, so hopefully this will catch potential issues)

packages/ios-client/scripts/lint_check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+echo "Running swift-format:check for ios files \n"
+swift-format format -i -r ./ --configuration swift-format-config.json
+swift-format lint -s -r ./ --configuration swift-format-config.json
+
```
MiloszFilimowski, line 7: **question:** What's the purpose of this being the same as lint.sh? 🤔

mironiasty: I think I just re-used convention. But it makes sense to remove duplicated file

MiloszFilimowski (approved):

PR #196, opened 2024-10-22
[FCE-725]: Standarize protobuf generation

MiloszFilimowski:
## Description

- Created a makefile that will fetch and use a locked version of protoc.
- Disabled Android linting 
- Updated submodule path to use ssh.

## Motivation and Context

- Generating protos was inconsistent between dev environments.

## How has this been tested?

- Checked the result of generated protos

## Types of changes

      not work as expected)

## Checklist:


packages/android-client/FishjamClient/src/main/java/com/protos/fishjam/PeerMessageKt.kt
MiloszFilimowski, line 1: **note:** Changes in this file are the result of disabled linting.

.gitignore
```diff
@@ -82,3 +82,7 @@ report.xml
 tsconfig.tsbuildinfo
 
 docs/
+
+# Proto
+.protoc/
+.swift-protobuf/
```
maksg, line 88: Please add a new line at the end in here, `.editorconfig` and `scripts/compile_protos.mk` to get rid of the warning

scripts/compile_protos.mk
```diff
@@ -0,0 +1,74 @@
+PROTOC_VERSION := 26.1
+SWIFT_PROTOBUF_VERSION := 1.25.2
```
mironiasty, line 2: I'm curious if there could be a way to have some automatic version update from time to time. As I'm afraid we will forget about these version for long time.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "protos"]
 	path = protos
-	url = https://github.com/fishjam-cloud/protos.git
+	url = git@github.com:fishjam-cloud/protos.git
```
mironiasty, line 3: Are you sure we wont this change? Our other repos use `https` in url

mironiasty (approved):

.gitmodules
MiloszFilimowski: as discussed, let's leave it with ssh version.

scripts/compile_protos.mk
MiloszFilimowski: I couldn't find a quick way to do that, however, I created a task to research if we can keep it in sync across SDK-s. 
https://linear.app/swmansion/issue/FCE-763

@PiotrWodecki suggested a docker image that could be shared.

PR #201, opened 2024-10-23
[FCE-683]: Update iOS WebRTC-SDK

MiloszFilimowski:
## Description

- Bumped WebRTC iOS SDK
- The whole issue was caused by this commit: https://github.com/webrtc-sdk/webrtc/pull/137 
If we ever need to support multi-cam, we might need to figure out how to make it work.

## Motivation and Context

WebRTC bump was needed for https://github.com/fishjam-cloud/mobile-client-sdk/pull/170

## How has this been tested?

- Tested if camera preview works on iOS.

## Types of changes

      not work as expected)

## Checklist:


mironiasty (approved):
Good catch! 👏 

>  we ever need to support multi-cam,

I don't expect this will be a case anytime soon. And if so, we will have few other issues anyway 🤷

PR #209, opened 2024-10-29
[FCE-799] Fix dependencies for RN package

mironiasty:
## Description

During investigation on using Expo 52, I found that we have incorrect project configuration. These PR removes some of unnecessary dependencies and fix linting issues.

## Motivation and Context

Let's make our configuration better. So once we decide to migrate, it will be easy ;)

## How has this been tested?

Make sure example app is still compiling and video is visible on devices.

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/scripts/lint_check.sh
```diff
@@ -14,6 +14,6 @@ cd ..
 
 echo "Running swift-format:check for react-native ios files \n"
 cd ios
-swift-format format -i -r ./ --configuration swift-format-config.json
 swift-format lint -s -r ./ --configuration swift-format-config.json
+swift-format format -i -r ./ --configuration swift-format-config.json
```
mironiasty, line 18: Fun fact: order in command makes a difference. If we first format and then lint -> formatting issues won't be catch on CI, as all files will be formatted, so linter won't spot issues in them. And script will return `0` return code (everything is ok)

packages/react-native-client/package.json
```diff
@@ -44,30 +44,28 @@
     "protobufjs": "^7.4.0"
   },
   "devDependencies": {
-    "@testing-library/react": "^15.0.7",
+    "@testing-library/dom": "^10.4.0",
```
mironiasty, line 47: This package is now required for `@testing-library/react` v16

MiloszFilimowski (approved):

PR #215, opened 2024-10-31
[FCE-804] Add branded IDs for peers and tracks

mironiasty:
## Description

Instead of strings, our IDs can be defined as branded types. This way it won't be possible* to use Track ID instead of Peer ID in typescript code

*ok, it would be possible if someone does not care about proper TS

## Motivation and Context

This change should make our API less error prone

## How has this been tested?

App is still passing CI checks

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
+++ b/packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
@@ -23,7 +23,6 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
     private var config: RTCConfiguration?
 
     private var midToTrackId: [String: String] = [:]
-    
 
     private static let mediaConstraints = RTCMediaConstraints(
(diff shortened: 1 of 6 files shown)


packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -23,7 +23,6 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
     private var config: RTCConfiguration?
 
     private var midToTrackId: [String: String] = [:]
-
```
mironiasty, line 26: This is just fix in formatting, that broke `main`.

MiloszFilimowski (approved):
Looks good, but keep in mind that this is a breaking change (although not very hard to fix). Users will get compilation errors after upgrading.

mironiasty:
It is rather small breaking change. You can still assign branded type to string. So in 90% of existing cases, this will not break a code.

PR #216, opened 2024-10-31
[FCE-764]: usePeers hook should have the same API as on web-sdk

MiloszFilimowski:
## Description

- Created the same API as on WEB. 
- Deprecated peers in docs
- Refactor existing code in example app to make it easier to read
- Create [web issue](https://linear.app/swmansion/issue/FCE-815) to also rename peers to remotePeers

## Motivation and Context

- Mobile was out of sync with Web.

## How has this been tested?

Run on Android/iOS

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -85,18 +98,100 @@ function addIsActiveToTracks<
...
+  ServerMetadata extends GenericMetadata = GenericMetadata,
+>(
+  peer: Peer<PeerMetadata, ServerMetadata>,
+): PeerWithTracks<PeerMetadata, ServerMetadata> {
+  const { tracks: peerTracks } = peer;
+  const distinguishedTracks: DistinguishedTracks = {};
+
+  for (const track of peerTracks) {
+    const trackType = track.metadata?.type;
+    if (!trackType) {
```
maksg, line 113: Would it work if we put this `continue` inside `default` case of the `switch` below instead? I think it might be cleaner that way

MiloszFilimowski: Good point. I can remove this "continue" altogether, it's not needed.

maksg (approved):

mironiasty (commented):
:ok:

PR #220, opened 2024-11-06
[FCE-806] Introduce connection hook

mironiasty:
## Description

This PR introduces new hook `useFishjamConnection`. It aims to replace useStatus*** hooks, so only one hook can be used in real apps. It also share `join/leaveRoom` methods.

## Motivation and Context

After some discussions with external developers,  it turned out that having static methods in our API is a little confusing in react world. So I decided to include them in hook, for better discoverability. Same hook will also share connections status. So for most cases, this will be only hook that has to be used to connect with Fishjam.

## How has this been tested?

This scenario was tested in example app. I does not change logic itself (or maybe a little), so most important part is how new API is structured.

## Types of changes

      not work as expected)

## Checklist:


maksg (approved):

maksg:
@mironiasty @MiloszFilimowski I was wondering if maybe we should change `useFishjamConnect` to `useConnectionStatus`. I'm changing documentation now and `useFishjamConnect` sounds a little out of place for me

mironiasty (commented):
LGTM
🚢 🇮🇹

PR #222, opened 2024-11-06
FCE-684: Simple component

MiloszFilimowski:
## Description

- Created the `FishjamRoom` component which should serve as a simple drop-in component for testing
- Created test tab for testing the above component.
- Refactored `startCamera` to return a boolean if the camera was started successfully. 

## Motivation and Context

A simple component was needed for a quicker user onboarding. 

## How has this been tested?

- Build, run and tested on iOS/Android. 

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/react-native-client/src/components/FishjamRoom/parsePeersToTracks.ts
+++ b/packages/react-native-client/src/components/FishjamRoom/parsePeersToTracks.ts
@@ -1,4 +1,4 @@
-import { PeerWithTracks } from '../..';
+import { PeerWithTracks } from '../../hooks/usePeers';
 import { GridTrack } from './GridTrackItem';
 
 const createGridTracksFromPeer = (peer: PeerWithTracks): GridTrack[] => {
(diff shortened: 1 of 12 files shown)


packages/react-native-client/src/components/FishjamRoom/parsePeersToTracks.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+  }
+
+  if (tracks.length === 0) {
+    tracks.push({
+      track: null,
+      peerId: peer.id,
+      isLocal: peer.isLocal,
+      isVadActive:
+        peer.microphoneTrack?.vadStatus === 'speech' ||
```
maksg, line 31: That's a lot of `speech` in this file. Maybe we could extract it to some variable instead of comparing directly to string?

MiloszFilimowski: This comes from `VadStatus` type so it's typed checked. 
`type VadStatus = 'silence' | 'speech';`

maksg (approved):

PR #223, opened 2024-11-06
[FCE-846] Make additional hooks stateless

maksg:
## Description

Refactored `useAppScreenShare`, `useScreenShare`, `usePeerStatus`, `useBandwithEstimation` code to use `useFishjamEventState` instead of `useState` and `useFishjamEvent`.

## Motivation and Context

The code will be simpler and less error-prone.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/react-native-client/src/hooks/usePeerStatus.ts
+++ b/packages/react-native-client/src/hooks/usePeerStatus.ts
@@ -1,6 +1,6 @@
-import { useCallback, useState } from 'react';
-import { ReceivableEvents, useFishjamEvent } from './useFishjamEvent';
+import { ReceivableEvents } from './useFishjamEvent';
 import RNFishjamClientModule from '../RNFishjamClientModule';
+import { useFishjamEventState } from './useFishjamEventState';
 
 /**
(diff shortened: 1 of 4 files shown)


packages/react-native-client/src/hooks/usePeerStatus.ts
```diff
@@ -19,15 +19,10 @@ export type PeerStatus = 'connecting' | 'connected' | 'error' | 'idle';
  * @group Hooks
  */
 export const usePeerStatus = () => {
-  const [peerStatus, setPeerStatus] = useState(
+  const peerStatus = useFishjamEventState<PeerStatus>(
+    ReceivableEvents.PeerStatusChanged,
```
MiloszFilimowski, line 23: **question:** It looks like PeerStatus was previously possible to be undefined. Can you confirm that it won't happen?

maksg: This value on native side is never optional. The only possibility for it to be undefined is if `PeerStatus` on the native side would have different value. I don't know if it's possible for us to desynchronise native and RN

MiloszFilimowski (approved):

PR #235, opened 2024-11-12
[FCE-853] Refactor useReconnection to be stateless

maksg:
## Description

Moved useReconnection status logic to native side.

## Motivation and Context

The code will be simpler and less error-prone.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
@@ -42,7 +42,7 @@ internal class ReconnectionManager(
       return
     }
 
-    if (reconnectionStatus == ReconnectionStatus.reconnecting) return
+    if (reconnectionStatus == ReconnectionStatus.reconnecting) { return }
     reconnectionStatus = ReconnectionStatus.reconnecting
 
     listeners.forEach { it.onReconnectionStarted() }
@@ -55,7 +55,7 @@ internal class ReconnectionManager(
   }
 
   fun onReconnected() {
-    if (reconnectionStatus != ReconnectionStatus.reconnecting) return
+    if (reconnectionStatus != ReconnectionStatus.reconnecting) { return }
     reset()
     listeners.forEach { it.onReconnected() }
   }
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
@@ -42,7 +42,9 @@ internal class ReconnectionManager(
       return
     }
 
-    if (reconnectionStatus == ReconnectionStatus.reconnecting) { return }
+    if (reconnectionStatus == ReconnectionStatus.reconnecting) {
+      return
+    }
     reconnectionStatus = ReconnectionStatus.reconnecting
 
     listeners.forEach { it.onReconnectionStarted() }
@@ -55,7 +57,9 @@ internal class ReconnectionManager(
   }
 
   fun onReconnected() {
-    if (reconnectionStatus != ReconnectionStatus.reconnecting) { return }
+    if (reconnectionStatus != ReconnectionStatus.reconnecting) {
+      return
+    }
     reset()
     listeners.forEach { it.onReconnected() }
   }
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
@@ -19,7 +19,9 @@ interface ReconnectionManagerListener {
   }
 }
 
-enum class ReconnectionStatus(val status: String) {
+enum class ReconnectionStatus(
+  val status: String
+) {
   Idle("idle"),
   Reconnecting("reconnecting"),
   Error("error")
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
@@ -19,11 +19,10 @@ interface ReconnectionManagerListener {
   }
 }
 
-@Suppress("ktlint:standard:enum-entry-name-case")
-enum class ReconnectionStatus {
-  idle,
-  reconnecting,
-  error
+enum class ReconnectionStatus(val status: String) {
+  Idle("idle"),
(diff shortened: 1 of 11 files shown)


packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -32,41 +32,37 @@ internal class ReconnectionManager(
...
-      reconnectionStatus = ReconnectionStatus.ERROR
+      reconnectionStatus = ReconnectionStatus.error
       listeners.forEach { it.onReconnectionRetriesLimitReached() }
       return
     }
 
-    if (reconnectionStatus == ReconnectionStatus.WAITING) {
-      return
-    }
+    if (reconnectionStatus == ReconnectionStatus.reconnecting) return
```
MiloszFilimowski, line 45: **question:** What do you think about always using brackets for if statements? It's easy to make an error accidentally and I think it doesn't really look that bad 😅

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -32,41 +32,37 @@ internal class ReconnectionManager(
...
+
     Timer().schedule(delay) {
-      reconnectionStatus = ReconnectionStatus.RECONNECTING
       connect()
     }
   }
 
   fun onReconnected() {
-    if (reconnectionStatus != ReconnectionStatus.RECONNECTING) return
+    if (reconnectionStatus != ReconnectionStatus.reconnecting) return
```
MiloszFilimowski, line 58: **question:** same question as above

MiloszFilimowski (approved):

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
maksg: I actually have no strong preferences for this. The if below didn't have brackets so I decided to make it consistent with less code. But I can add the brackets if only because Swift requires those brackets!

packages/ios-client/Sources/FishjamClient/ReconnectionManager.swift
```diff
@@ -39,43 +38,32 @@ class ReconnectionManager {
...
+        guard reconnectAttempts < reconnectConfig.maxAttempts else {
+            reconnectionStatus = .error
             listener.onReconnectionRetriesLimitReached()
             return
         }
 
-        if reconnectionStatus == .WAITING {
-            return
-        }
+        guard reconnectionStatus != .reconnecting else { return }
```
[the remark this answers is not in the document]
maksg: I could kind of agree that it might be easier to read. However I don't think that it's that much of an issue and for me guards are much easier to spot and by extension the whole block becomes easier to understand. `if return`s just look strange for me but that could just be bacause of my long time of working with Swift 😅

packages/react-native-client/src/hooks/useFishjamEvent.ts
```diff
@@ -13,6 +13,7 @@ export const ReceivableEvents = {
   ReconnectionRetriesLimitReached: 'ReconnectionRetriesLimitReached',
   ReconnectionStarted: 'ReconnectionStarted',
   Reconnected: 'Reconnected',
```
mironiasty, line 15: Can these values be removed now?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
```diff
@@ -19,11 +19,11 @@ interface ReconnectionManagerListener {
...
+@Suppress("ktlint:standard:enum-entry-name-case")
 enum class ReconnectionStatus {
-  IDLE,
-  RECONNECTING,
-  ERROR,
-  WAITING
+  idle,
+  reconnecting,
+  error
 }
```
mironiasty, line 27: WDYT about sth like:
```
enum class ReconnectionStatus(val status: String) {
  IDLE("idle"),
  RECONNECTING("reconnecting"),
  ERROR("error")
}
```

packages/react-native-client/src/hooks/useReconnection.ts
```diff
@@ -10,21 +11,9 @@ export type ReconnectionStatus = 'idle' | 'reconnecting' | 'error';
...
-  const setStatusError = useCallback(() => setReconnectionStatus('error'), []);
-
-  useFishjamEvent(ReceivableEvents.Reconnected, setStatusIdle);
-  useFishjamEvent(ReceivableEvents.ReconnectionStarted, setStatusReconnecting);
-  useFishjamEvent(
-    ReceivableEvents.ReconnectionRetriesLimitReached,
-    setStatusError,
+  const reconnectionStatus = useFishjamEventState<ReconnectionStatus>(
+    ReceivableEvents.ReconnectionStatusChanged,
+    RNFishjamClientModule.reconnectionStatus,
```
mironiasty, line 16: I wonder if we should consider some internal additional guards to protect against receiving incorrect values from native part. LIke some debug-time checking if `RNFishjamClientModule.reconnectionStatus` will have only expected values 🤔 (but this is not for this PR)

cc @MiloszFilimowski

packages/react-native-client/ios/Events.swift
```diff
@@ -20,6 +20,7 @@ class EmitableEvent {
         case ReconnectionRetriesLimitReached
         case ReconnectionStarted
         case Reconnected
```
mironiasty, line 22: Is it still in use?

packages/react-native-client/src/hooks/useFishjamEvent.ts
maksg: It depends. Do we want to remove those events for users completely? I left them in case someone relies on them and doesn't want to switch to reconnection status

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/ReconnectionManager.kt
maksg: All caps enum values are discouraged in Kotlin. I did this change so it looks better while being simpler and because PeerStatus was already done this way.

However it's true that first letters should be uppercased if I want to follow Kotlin rules.
I can change this to:
```
enum class ReconnectionStatus(val status: String) {
  Idle("idle"),
  Reconnecting("reconnecting"),
  Error("error")
}
```

packages/react-native-client/src/hooks/useReconnection.ts
maksg: Agreed. This also came up during `usePeerStatus` although we decided that it's not an issue at the moment

PR #246, opened 2024-11-19
FCE-748: support for protobufs

MiloszFilimowski:
## Description

- Updated mobile SDK to support protobufs for media events.
- This PR does not support old version of handling media events via JSON so this is a **BREAKING CHANGE.**

## Motivation and Context

New version of fishjam added protobufs for media events.

## How has this been tested?

- Run and tested bidirectional connection between Android/iOS

## Types of changes

      not work as expected)


packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/events/Event.kt
```diff
@@ -1,429 +1,15 @@
...
-
-          else ->
-            null
-        }
-      } catch (e: JsonParseException) {
-        Timber.e(e)
-        return null
-      }
-    }
+internal fun String.serializeToMap(): Map<String, Any?> =
```
maksg, line 9: This file only has this one function and it's not an event so maybe we should consider changing the filename?

MiloszFilimowski: Good point, it used to have Events but now they are gone. I forgot to rename the file.

packages/ios-client/Sources/FishjamClient/utils/AnyJson.swift
```diff
@@ -230,3 +236,13 @@ public struct AnyJson: Codable {
         var stringValue: String { return key }
     }
 }
+
+extension String {
```
maksg, line 240: Are those really better than just using `AnyJson(from: )`? Seems a little bit like an overkill

packages/ios-client/Sources/FishjamClient/utils/AnyJson.swift
```diff
@@ -230,3 +236,13 @@ public struct AnyJson: Codable {
...
+    func toAnyJson() throws -> AnyJson {
+        return try AnyJson(from: self)
+    }
+
+    func toAnyJson() -> AnyJson? {
+        return try? AnyJson(from: self)
```
maksg, line 246: And if they are better then this could be `try? toAnyJson()`

packages/ios-client/Sources/FishjamClient/utils/types.swift
```diff
@@ -21,3 +23,32 @@ extension AnyJson {
         return res
     }
 }
+
+public struct JsonEncodingError: Error {}
+
+extension Encodable {
+
+    public func toJsonString() throws -> String {
+        if let json = String(data: try JSONEncoder().encode(self), encoding: .utf8) {
```
maksg, line 32: I know it's controversial but what about `guard` and `throw` in `else` instead?

packages/ios-client/Sources/FishjamClient/utils/types.swift
```diff
@@ -21,3 +23,32 @@ extension AnyJson {
...
+            return json
+        }
+        throw JsonEncodingError()
+    }
+
+    var toJsonStringOrEmpty: String {
+        do {
+            return try self.toJsonString()
```
maksg, line 40: We can remove `self` here

packages/ios-client/Sources/FishjamClient/utils/types.swift
```diff
@@ -21,3 +23,32 @@ extension AnyJson {
...
+        } catch {
+            sdkLogger.log(level: .error, "Unable to encode metadata")
+            return "{}"
+        }
+    }
+}
+
+extension [String: Metadata] {
+    public func toDictionaryJson() -> [String: String] {
+        return mapValues { val in
```
maksg, line 50: I think `mapValues(\.toJsonStringOrEmpty)` should work here

packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -94,14 +94,18 @@ function addIsActiveToTracks<
...
-  return peers.map((peer) => ({
-    ...peer,
-    tracks: peer.tracks.map((track) => ({
-      ...track,
-      isActive:
-        (track as { metadata?: TrackMetadata })?.metadata?.active ?? true,
-    })),
-  }));
+  return peers.map((peer) => {
+    return {
```
maksg, line 98: Remove `return`s

maksg (changes requested):

maksg:
I left some comments. But other than that great job! 🎉

packages/ios-client/Sources/FishjamClient/utils/AnyJson.swift
MiloszFilimowski: Fair point, I just thought it looks better but you're right.

packages/ios-client/Sources/FishjamClient/utils/types.swift
maksg: NO SHOUTING IN THE CHAT PLEASE, PEOPLE ARE TRYING TO WORK HERE

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -176,11 +176,9 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
         config.candidateNetworkPolicy = .all
         config.tcpCandidatePolicy = .disabled
 
-        // if ice servers are not empty that probably means we are using turn servers
+        // if ice servers are not empty that probably means we are using turn servers, which are not handled at the moment
```
roznawsk, line 179: **suggestion:** 

We might actually start using turn servers, that will be passed in the `connected` message. They will be handled out of the box
```suggestion
```

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -176,11 +176,9 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
...
         if iceServers.count > 0 {
```
roznawsk, line 180: **suggestion:** 
If they are empty, then we can just pass this empty list from protobuf message
```suggestion
```

packages/ios-client/Sources/FishjamClient/webrtc/PeerConnectionManager.swift
```diff
@@ -308,6 +273,22 @@ internal class PeerConnectionManager: NSObject, RTCPeerConnectionDelegate {
...
+        guard let pc = connection else {
+            return [:]
+        }
+
+        var mapping: [String: Int32] = [:]
+        pc.transceivers.forEach { transceiver in
+            guard let trackId: String = transceiver.sender.track?.trackId else {
+                return
+            }
+            mapping[trackId] = 500  // TODO: Change with simulcast
```
roznawsk, line 286: **suggestion:** more appropriate value would be `1500000` as this value is in bit\s

packages/ios-client/Sources/FishjamClient/webrtc/RTCEngineCommunication.swift
```diff
@@ -17,150 +17,167 @@ internal class RTCEngineCommunication {
...
+        sdpOffer.sdp = sdp
+        sdpOffer.trackIDToMetadataJson = trackIdToTrackMetadata.toDictionaryJson()
+        sdpOffer.midToTrackID = midToTrackId
+        sdpOffer.trackIDToBitrates = Dictionary(
+            uniqueKeysWithValues:
+                trackIdToBitrates.map { key, value in
+                    var trackBitrates = Fishjam_MediaEvents_Peer_MediaEvent.TrackBitrates()
+                    trackBitrates.trackID = key
+                    var bitrate = Fishjam_MediaEvents_Peer_MediaEvent.VariantBitrate()
+                    bitrate.variant = .unspecified  // TODO: Add with simulcast
```
roznawsk, line 75: **note:** this will remain .unspecified for non-simulcast tracks

packages/ios-client/Sources/FishjamClient/webrtc/RTCEngineCommunication.swift
```diff
@@ -17,150 +17,167 @@ internal class RTCEngineCommunication {
...
     }
 
     func localCandidate(sdp: String, sdpMLineIndex: Int32, sdpMid: Int32, usernameFragment: String) {
-        sendEvent(
-            event: LocalCandidateEvent(
-                candidate: sdp, sdpMLineIndex: sdpMLineIndex, sdpMid: sdpMid, usernameFragment: usernameFragment))
+        var candidate = Fishjam_MediaEvents_Candidate()
+        candidate.candidate = sdp
+        candidate.sdpMLineIndex = sdpMLineIndex
+        candidate.sdpMid = String(sdpMid)
```
roznawsk, line 54: **praise:** 👍🏽

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/PeerConnectionManager.kt
```diff
@@ -400,7 +379,7 @@ internal class PeerConnectionManager(
     peerConnectionMutex.withLock {
       val pc = peerConnection!!
 
-      if (needsRestart && !isExWebrtc) {
+      if (needsRestart) {
         pc.restartIce()
       }
```
roznawsk, line 384: **suggestion:** In web-sdk we removed this completely. We assume ice restarts are not needed during negotiation. In fact in web-sdk the restarts introduced failures. Try removing this and checking if everything works **OK**
```suggestion
```

roznawsk (changes requested):

examples/fishjam-chat/plugin/src/with-custom-config-ios.ts
```diff
@@ -13,9 +13,9 @@ function replaceCloudClientForExtension(podfileContent: string) {
     'g',
   );
 
-  podfileContent = podfileContent.replace(targetRegex, (match) => {
-    return match.replace(podToReplace, replacementPod);
-  });
+  podfileContent = podfileContent.replace(targetRegex, (match) =>
```
mironiasty, line 16: I think this new rule could be decoupled from this PR... Bet let it be as is.

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -162,6 +161,7 @@ internal class FishjamClientInternal(
                 PeerNotifications.PeerMessage.AuthRequest
                   .newBuilder()
                   .setToken(connectConfig.token)
+                  .setSdkVersion("mobile-0.6.0")
```
mironiasty, line 164: Can we read this value from config files?

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -621,23 +630,20 @@ internal class FishjamClientInternal(
...
             tracksTypes,
             localEndpoint.tracks.values.toList()
           )
 
         rtcEngineCommunication.sdpOffer(
           offer.description,
           localEndpoint.tracks.map { (_, track) -> track.webrtcId() to track.metadata }.toMap(),
-          offer.midToTrackIdMapping
+          offer.midToTrackIdMapping,
+          localEndpoint.tracks.map { (_, track) -> track.webrtcId() to 1500000 }.toMap() // TODO: Update with simulcast
```
mironiasty, line 646: WDYT about adding link to simulcast project in linear?
```suggestion
          localEndpoint.tracks.map { (_, track) -> track.webrtcId() to 1500000 }.toMap() // TODO(FCE-XXX): Update with simulcast
```

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/webrtc/RTCEngineCommunication.kt
```diff
@@ -44,34 +20,69 @@ internal class RTCEngineCommunication {
...
     trackId: String,
     encoding: TrackEncoding
   ) {
-    sendEvent(
-      SelectEncoding(
-        trackId,
-        encoding.rid
-      )
-    )
+    // TODO: This will be useful after simulcast is enabled
```
mironiasty, line 72:
```suggestion
    // TODO(FCE-XXX): This will be useful after simulcast is enabled
```

packages/android-client/FishjamClient/src/main/java/fishjam/media_events/CandidateKt.kt
```diff
@@ -0,0 +1,106 @@
+// Generated by the protocol buffer compiler. DO NOT EDIT!
```
mironiasty, line 1: idea for future: would it be possible to keep this generated files in folder called `generated`?

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
```diff
@@ -345,14 +345,15 @@ class FishjamClientInternal {
...
-        rtcEngineCommunication.onEvent(serializedEvent: event)
+    private func receiveEvent(event: Fishjam_MediaEvents_Server_MediaEvent) {
+        rtcEngineCommunication.onEvent(event: event)
     }
 
     func websocketDidConnect() {
         let authRequest = Fishjam_PeerMessage.with({
             $0.authRequest = Fishjam_PeerMessage.AuthRequest.with({
                 $0.token = self.config?.token ?? ""
+                $0.sdkVersion = "mobile-0.6.0"
```
mironiasty, line 356: Could it be read from config file?

mironiasty (commented):
Two general comments: 
- can we read package version from configuration file, so releasing app will require less work with updating?
- I would consider adding link to linear item for simulcast - it might help in future, to understand todo in code

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
MiloszFilimowski: 🫡 https://linear.app/swmansion/issue/FCE-952

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
MiloszFilimowski: 🫡 https://linear.app/swmansion/issue/FCE-952

packages/android-client/FishjamClient/src/main/java/fishjam/media_events/CandidateKt.kt
MiloszFilimowski: Unfortunately it looks like it's not possible, as the package name comes from the protofile and `protoc` does not provide the ability to modify it. So although we could change the file location it won't work on Android as it expects certain folder structure. 

We could modify the protos file, but this might destroy code generation for others...

roznawsk (approved):

maksg (approved):

packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
mironiasty: Thanks!
I also updated release instruction to keep it in mind in upcoming release
https://github.com/fishjam-cloud/internal-docs/wiki/Mobile-SDK-release

mironiasty (approved):
🥔

PR #253, opened 2024-11-22
[FCE-891] Add camera and microphone usage descriptions and remove `patch-package`

maksg:
## Description

- Added `NSCameraUsageDescription` and `NSMicrophoneUsageDescription` to `Info.plist`.
- Added patch which removes push notifications code from `expo-modules-core`.

Push notifications issue may be resolved in future Expo SDKs so we should keep tracking this issue in case they provide a better solution.

This fix has been taken from: https://github.com/expo/fyi/blob/main/apns-entitlement-sdk-51.md

## Motivation and Context

When uploading build to App Store there were warnings regarding those two things. Additionally missing usage descriptions prevented the build from appearing in App Store Connect. 

## Types of changes

      not work as expected)

## Checklist:


package.json
```diff
@@ -21,7 +21,7 @@
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-prettier": "^5.2.1",
     "eslint-plugin-react-hooks": "^5.0.0",
-    "patch-package": "8.0.0",
+    "patch-package": "^8.0.0",
```
MiloszFilimowski, line 24: **question:** why this change?

package.json
```diff
@@ -34,6 +34,10 @@
     "lint": "yarn workspaces foreach -A -p run lint && yarn prettier --write .github/*",
     "lint:check": "yarn workspaces foreach -A -p run lint:check && yarn prettier --check .github/*",
     "docs": "typedoc",
-    "gen:proto": "make -f ./scripts/compile_protos.mk"
+    "gen:proto": "make -f ./scripts/compile_protos.mk",
+    "postinstall": "patch-package"
```
MiloszFilimowski, line 38: **issue:** Is it possible to have it in package.json for `examples/fishjam-chat`? It's not something that concerns our SDK 🤔

package.json
maksg: It happened automatically when I installed `patch-package`. I didn't notice that it was already installed here. I can revert that if you wish

package.json
maksg: I tried adding it to `examples/fishjam-chat` but unfortunately it does not work there. The `node_modules` are in the root directory and `patch-package` can't find them from there. I didn't find a way to force it to look for that elsewhere but I'm open for suggestions

MiloszFilimowski: What do you think about this feature? 🤔 
https://yarnpkg.com/cli/patch

mironiasty (approved):
LGTM

MiloszFilimowski (approved):

PR #261, opened 2024-11-28
[FCE-931] Unify joinRoom signature with web

mironiasty:
## Description

This is change in `joinRoom` API where it will be unified with web API - param will be object.

Here is related change in web SDK: https://github.com/fishjam-cloud/web-client-sdk/pull/206
Here is related change in documentation: https://github.com/fishjam-cloud/documentation/pull/62

## Motivation and Context

We should aim for similar API on all platforms, and this is yet another step to unify it (this time Web part had better approach)

## How has this been tested?

This is 'just' a change in method typing.

## Types of changes

      not work as expected)

## Checklist:


MiloszFilimowski (approved):
LGTM

PR #278, opened 2024-12-12
[FCE-876]: Create test that verify if native events have expected value

MiloszFilimowski:
## Description

- Added TS types for payloads 
- Added runtime dev only validation to check if events sent from the native side match with our expectations. 

## Motivation and Context

- We want to make sure events sent from the native side have expected values. 

## How has this been tested?

- Tested if error is thrown when unexpected value happens.

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/src/hooks/usePeers.ts
```diff
@@ -166,10 +166,14 @@ export function usePeers<
...
-  const peers = useFishjamEventState<Peer<PeerMetadata, ServerMetadata>[]>(
+  const peers = useFishjamEventState(
     ReceivableEvents.PeersUpdate,
     RNFishjamClientModule.getPeers<PeerMetadata, ServerMetadata>(),
-    (peersWithoutActive) => addIsActiveToTracks(peersWithoutActive),
+    (peersWithoutActive) =>
+      addIsActiveToTracks(peersWithoutActive) as Peer<
+        PeerMetadata,
+        ServerMetadata
+      >[],
```
mironiasty, line 176: Super nit: would it be cleaner to add casting to `useFishjamEventState` resutl? (instead of casting 3rd param)

packages/react-native-client/src/debug/internal/eventPayloadValidator.ts
mironiasty, line 1: I think this file could go to `src/utils` directory 🤔

packages/react-native-client/src/debug/internal/eventPayloadValidator.ts
```diff
@@ -0,0 +1,226 @@
...
+import type {
+  Camera,
+  CameraFacingDirection,
+  CurrentCameraChangedType,
+} from '../../hooks/useCamera';
+import type { AudioTrack, Peer, VideoTrack } from '../../hooks/usePeers';
+import type { ReconnectionStatus } from '../../hooks/useConnection';
+import type { PeerStatus } from '../../hooks/useConnection';
+
+export function isNativeEventPayloadValid<
```
mironiasty, line 25: WDYT about using [zod](https://zod.dev) here? Even just for fun.

packages/react-native-client/src/debug/internal/eventPayloadValidator.ts
```diff
@@ -0,0 +1,226 @@
...
+  if (!value || typeof value !== 'object') {
+    return false;
+  }
+  const config = value as SimulcastConfig;
+  const validEncodings = ['l', 'm', 'h'];
+  return (
+    typeof config.enabled === 'boolean' &&
+    Array.isArray(config.activeEncodings) &&
+    config.activeEncodings.every(
+      (e) => typeof e === 'string' && validEncodings.includes(e),
```
mironiasty, line 92: `e` usually means `error` in JS. So maybe `encoding`?

mironiasty (approved):

packages/react-native-client/src/hooks/usePeers.ts
MiloszFilimowski: This cast is actually not needed at all, I'm not sure why it's here. Thanks!

packages/react-native-client/src/debug/internal/eventPayloadValidator.ts
MiloszFilimowski: Added!

mironiasty (approved):

packages/react-native-client/src/utils/eventPayloadValidator.ts
```diff
@@ -0,0 +1,120 @@
...
+  name: z.string(),
+  type: z.enum(['bluetooth', 'headset', 'speaker', 'earpiece']),
+});
+
+const AudioDeviceEventSchema = z.object({
+  availableDevices: z.array(AudioOutputDeviceSchema),
+  selectedDevice: AudioOutputDeviceSchema.nullable(),
+});
+
+const CameraSchema = z.object({
```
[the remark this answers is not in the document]
MiloszFilimowski: We would then have to replace all of our types to use zod to be consistent. Do you think it's worth it?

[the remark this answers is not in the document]
mironiasty: Ok, let's leave it as is

PR #285, opened 2025-01-15
Release 0.7.0

MiloszFilimowski:
Release 0.7.0

### DIFF
--- a/examples/fishjam-chat/package.json
+++ b/examples/fishjam-chat/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "main": "index.js",
   "scripts": {
-    "eas-build-pre-install": "yarn workspace fishjam-mobile-sdk install && yarn workspace fishjam-mobile-sdk build",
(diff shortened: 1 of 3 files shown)


examples/fishjam-chat/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "main": "index.js",
   "scripts": {
-    "eas-build-pre-install": "yarn workspace fishjam-mobile-sdk install && yarn workspace fishjam-mobile-sdk build",
+    "eas-build-pre-install": "corepack enable && yarn workspace fishjam-mobile-sdk install && yarn workspace fishjam-mobile-sdk build",
```
mironiasty, line 6: (might be done as separate task)

WDYT about having separate script for these commands? 🤔

mironiasty (approved):
👍

examples/fishjam-chat/package.json
MiloszFilimowski: It must be like that, this is the only script that eas runs before building.

czerwiukk (approved):

PR #322, opened 2025-02-06
FCE-1179: Emit a warning when codec not supported

MiloszFilimowski:
## Description

- Added a warning on Android when H264 is not supported by device

## Motivation and Context

- Right now the stream just won't work which might confuse users.

## How has this been tested?

- Tested on Android emulator and a real device.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
+++ b/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
@@ -927,4 +927,13 @@ class RNFishjamClient(
       )
     )
   }
+
+  override fun onCodecsOffered(codecs: List<String>) {
+    // TODO: Remove when ex_webrtc adds support for both VP8 and H264
+    // Right now all rooms on Sandbox are created with H264. Android emulators
+    // usually won't support this, so let's inform users for now why stream is not working.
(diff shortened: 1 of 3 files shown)


packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -927,4 +927,13 @@ class RNFishjamClient(
       )
     )
   }
+
+  override fun onCodecsOffered(codecs: List<String>) {
+    // TODO: Remove when ex_webrtc adds support for both VP8 and H264
+    // Right now all rooms on Sandbox are created with H264. Android emulators
+    // usually won't support this, so let's inform users for now why stream is not working.
+    if (!codecs.contains("H264")) {
+      emitEvent(EmitableEvent.warning("H264 codec is not supported on your device. Video won't be sent/received."))
```
mironiasty, line 936: 🤔 
I think it might be not helpful to end users. If I'd get such warning, I would have no idea what to do next.

Maybe we should create a docs page that describes this issue with explanation and workarounds. And then put a link in the warning?

MiloszFilimowski:
Closing as this is not a viable solution.

PR #333, opened 2025-02-18
FCE-1188: Simplify RN FishjamRoom example

MiloszFilimowski:
## Description

- Introduced FishjamSandboxRoom which allows to connect directly to Sandbox using room manager. 

## Motivation and Context

- Our current minimal example is too complex. A single component with minimal properties provided was needed. 

## How has this been tested?

- Run and tested on Android.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/examples/fishjam-chat/screens/ConnectWithFishjamRoom.tsx
+++ b/examples/fishjam-chat/screens/ConnectWithFishjamRoom.tsx
@@ -1,46 +1,9 @@
-import { FishjamRoom } from '@fishjam-cloud/react-native-client';
-import { useEffect, useState } from 'react';
-import { joinRoomWithRoomManager } from '../utils/roomManager';
-import { ActivityIndicator, StyleSheet } from 'react-native';
+import { FishjamSandboxRoom } from '@fishjam-cloud/react-native-client';
 
-type RoomData = {
-  url: string;
-  peerToken: string;
-};
-
-export const ConnectWithFishjamRoom = () => {
-  const [roomData, setRoomData] = useState<RoomData | null>(null);
-
-  useEffect(() => {
-    const fetchRoomData = async () => {
-      try {
-        const { fishjamUrl, token } = await joinRoomWithRoomManager(
-          'https://room.fishjam.work/api/rooms',
-          'test-room',
-          'test-user',
-        );
-        setRoomData({
-          url: fishjamUrl,
-          peerToken: token,
-        });
-      } catch (_) {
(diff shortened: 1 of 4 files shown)


examples/fishjam-chat/screens/ConnectWithFishjamRoom.tsx
```diff
@@ -1,46 +1,9 @@
...
-
-const styles = StyleSheet.create({
-  indicator: {
-    flex: 1,
-  },
-});
+import { FishjamSandboxRoom } from '@fishjam-cloud/react-native-client';
+
+export const ConnectWithFishjamRoom = () => (
+  <FishjamSandboxRoom
```
MiloszFilimowski, line 4: **note:** I'm not emotionally attached to this name. So if somebody has a better idea, let's change it!

packages/react-native-client/src/components/FishjamSandboxRoom.tsx
```diff
@@ -0,0 +1,112 @@
+import { useEffect, useState } from 'react';
```
mironiasty, line 1: I'm not sure if this should be indeed part of SDK 🤔 
I know this would simplify examples, but it would also move part of example app into SDK. And this code can't be used in production app anyway.

mironiasty: Maybe such component can be part of example app?

packages/react-native-client/src/components/FishjamRoom/index.tsx
```diff
@@ -30,7 +32,11 @@ export type FishjamRoomProps = {
  * @param {string} props.fishjamUrl
  * @param {string} props.peerToken
  */
-export const FishjamRoom = ({ fishjamUrl, peerToken }: FishjamRoomProps) => {
+export const FishjamRoom = ({
+  fishjamUrl,
+  peerToken,
+  style,
```
mironiasty, line 38: Maybe `containerStyle` would be better name here? 🤔

packages/react-native-client/src/components/FishjamSandboxRoom.tsx
MiloszFilimowski: I agree, it might not fit perfectly, but it's not so bad to have this if it simplifies it for our users. If we add it to the example it won't be possible to import it so it defeats the purpose. The key here is to be able to easily test it inside your app.

PR #351, opened 2025-03-14
FCE-1318: Screenshare changes

MiloszFilimowski:
## Description

- Introduces CustomSource for clients to be able to attach any external source that can output CMSampleBuffer
- Refactored existing code to make it more extensible

## Motivation and Context

- This change was needed to add screen sharing on MacOS Catalyst.
- If we could implement a similar thing on Android, this would allow us to support external Camera Captureres and will resolve FCE-1290
- Since this API is only available for iOS, let's not document it until Android is figured out.

## How has this been tested?

- Successfully connected external source, streamed and displayed. 

## Types of changes

      not work as expected)

## Checklist:


packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
```diff
@@ -168,13 +170,49 @@ class FishjamClientInternal {
         return audioTrack
     }
 
+    public func create(customSource: CustomSource) async throws {
```
mironiasty, line 173: WDYT about less generic name? This class already has some `createxxx` methods

packages/ios-client/Sources/FishjamClient/media/Tracks/LocalAppScreenShareTrack.swift
```diff
@@ -1,27 +1,19 @@
...
-        self.videoSource = videoSource
-        super.init(mediaTrack: mediaTrack, endpointId: endpointId, rtcEngineId: nil, metadata: metadata)
-    }
-
-    convenience init(mediaTrack: RTCVideoTrack, oldTrack: LocalAppScreenShareTrack) {
-        self.init(
-            mediaTrack: mediaTrack, videoSource: oldTrack.videoSource, endpointId: oldTrack.endpointId,
-            metadata: oldTrack.metadata, videoParameters: oldTrack.videoParameters, capturer: oldTrack.capturer)
+        super.init(
+            mediaTrack: mediaTrack, videoSource: videoSource, endpointId: endpointId, rtcEngineId: nil,
```
mironiasty, line 15: I was surprised that linter allowed such syntax. But looks like we could enable `lineBreakBeforeEachArgument` rule in `swift-format-config.json` in another PR 🤔

mironiasty (approved):

packages/ios-client/Sources/FishjamClient/FishjamClientInternal.swift
MiloszFilimowski: This is a very common convention in Swift, the name of this function is createCustomSource, but because of the convention, it's split into the first argument.

PR #352, opened 2025-03-17
FCE-1328: Fix release action envs

MiloszFilimowski:
## Description

- Transformed release rn to action
- npm token is now required and correctly passed from workflows

## Motivation and Context

- the workflow was failing when called from a different workflow because of missing npm token

## How has this been tested?

- not possible to test until next release

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/.github/workflows/publish-react-native.yml
+++ b/.github/workflows/publish-react-native.yml
@@ -1,7 +1,6 @@
 # If for some reason release.yml fails for iOS or Android, but the spec is public, you can try to publish the React Native package manually.
 name: Publish React Native Package
 on:
-  workflow_call:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


czerwiukk (approved):
looks good

.github/workflows/publish-react-native.yml
```diff
@@ -12,11 +11,7 @@ jobs:
       id-token: write
     steps:
       - uses: actions/checkout@v4
```
czerwiukk, line 13: **question:** couldn't checkout be also included in the extracted workflow?

mironiasty (approved):

.github/workflows/publish-react-native.yml
MiloszFilimowski: It's extracted as an action, so an action should be independent of other steps IMO

PR #362, opened 2025-03-30
FCE-1389: Create CustomSource for Android

MiloszFilimowski:
## Description

- Created CustomSource for Android

## Motivation and Context

This PR will allow us to implement custom sources (like react-native-vision-camera) for both Android and iOS.

## How has this been tested?

- Not tested yet as there is no example, tested only with sample code on Android.

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/android-client/FishjamClient/build.gradle
+++ b/packages/android-client/FishjamClient/build.gradle
@@ -108,6 +108,7 @@ dependencies {
     implementation 'androidx.core:core:1.8.20'
 
     api 'io.github.webrtc-sdk:android:125.6422.06'
+    compileOnly 'androidx.camera:camera-core:1.5.0-alpha03'
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation "io.mockk:mockk:1.13.2"
--- a/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
+++ b/packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
@@ -9,6 +9,7 @@ import com.fishjamcloud.client.media.RemoteAudioTrack
 import com.fishjamcloud.client.media.RemoteVideoTrack
 import com.fishjamcloud.client.media.Track
 import com.fishjamcloud.client.models.AuthError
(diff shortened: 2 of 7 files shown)


packages/android-client/FishjamClient/src/main/java/com/fishjamcloud/client/FishjamClientInternal.kt
```diff
@@ -72,6 +74,8 @@ internal class FishjamClientInternal(
 
   private lateinit var reconnectionManager: ReconnectionManager
 
+  private val customSourcesManager = CustomSourceManager()
```
mironiasty, line 77: nit: why `Sources` vs `Source`? 🤔

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -517,6 +523,14 @@ class RNFishjamClient(
                   "metadata" to track.metadata
                 )
 
+              is VideoTrack ->
```
mironiasty, line 526: Does exposing new type of video track will not break anything in existing apps? 🤔

packages/android-client/FishjamClient/build.gradle
```diff
@@ -108,6 +108,7 @@ dependencies {
     implementation 'androidx.core:core:1.8.20'
 
     api 'io.github.webrtc-sdk:android:125.6422.06'
+    compileOnly 'androidx.camera:camera-core:1.5.0-alpha03'
```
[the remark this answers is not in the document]
MiloszFilimowski: I used the same version as rn-vision-camera uses, ideally this won't be needed in the future (I'd like to remove ImageProxy dependency) but for now as it's our only use case I think it might be easier.

packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
MiloszFilimowski: It should not as other cases should be matched before it. It is linked with the todo above. After LocalVideoTrack is refactored, this can be removed. But as with other cases, I want to cleanup after I have a working example ready.

PR #445, opened 2025-07-08
Fce 1717  configure plugins

dawidmatyjasik:
## Description

We added ability to make plugin more flexible

## How has this been tested?

on simulators 

## Types of changes

      not work as expected)

## Checklist:


packages/react-native-client/plugin/src/types.ts
```diff
@@ -8,6 +8,9 @@ export type FishjamPluginOptions =
         iphoneDeploymentTarget?: string;
         enableScreensharing?: boolean;
         supportsPictureInPicture?: boolean;
+        appGroupName?: string;
```
MiloszFilimowski, line 11: **suggestion:** I think it should be called `appGroupContainerId` to be more in sync with apple docs: https://developer.apple.com/documentation/xcode/configuring-app-groups

examples/minimal-react-native/app.json
```diff
@@ -35,6 +35,16 @@
...
     "web": {
       "favicon": "./assets/favicon.png"
-    }
+    },
+    "plugins": [
+      [
+        "@fishjam-cloud/react-native-client",
+        {
+          "ios": {
+            "mainTarget": "reactnativetest"
```
MiloszFilimowski, line 44: **question:** was this intended to stay here?

dawidmatyjasik: good catch!

MiloszFilimowski (approved):

PR #474, opened 2025-08-22
FCE-1882: Fix multiple calls to joinRoom

MiloszFilimowski:
## Description

- Emit a warning and return when `joinRoom` is called after already connected/connecting.
- Bumped rn-whip-whep (broken release was causing iOS build to fail)

## Motivation and Context

- Calling `joinRoom` multiple times was causing an undefined state, which was difficult to debug. 

## How has this been tested?

iOS + Android

## Types of changes

      not work as expected)

## Checklist:

### DIFF
--- a/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
+++ b/packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
@@ -299,7 +299,7 @@ class RNFishjamClient(
     promise: Promise
   ) {
     if (connectPromise != null || peerStatus == PeerStatus.Connected) {
-      emitEvent(EmitableEvent.warning("Room already joined or it's connecting. You must call leaveRoom before calling joinRoom again."))
+      emitEvent(EmitableEvent.warning("Room already joined or it's connecting. You must call leaveRoom() before calling joinRoom() again."))
       return
     }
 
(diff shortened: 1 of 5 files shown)


packages/react-native-client/android/src/main/java/io/fishjam/reactnative/RNFishjamClient.kt
```diff
@@ -298,6 +298,11 @@ class RNFishjamClient(
     config: ConnectConfig,
     promise: Promise
   ) {
+    if (connectPromise != null || peerStatus == PeerStatus.Connected) {
+      emitEvent(EmitableEvent.warning("Room already joined or it's connecting. You must call leaveRoom before calling joinRoom again."))
```
PiotrWodecki, line 302: **suggestion (non-blocking):** 
```suggestion
      emitEvent(EmitableEvent.warning("Room already joined or it's connecting. You must call leaveRoom() before calling joinRoom() again."))
```

packages/react-native-client/ios/RNFishjamClient.swift
```diff
@@ -240,6 +245,8 @@ class RNFishjamClient: FishjamClientListener {
     }
 
     func leaveRoom() {
+        connectPromise?.reject("E_MEMBRANE_CONNECT", "leaveRoom called before connected")
```
PiotrWodecki, line 248: **question (non-blocking):** What's E_MEMBRANE_CONNECT?

PiotrWodecki (approved):

packages/react-native-client/ios/RNFishjamClient.swift
MiloszFilimowski: This is just a legacy name of the exception. Won't be displayed to the users.

PR #479, opened 2025-09-02
Make FishjamRoom use id instead of url to join room

anna1901:
## Description

Change the implementation of FishjamRoom so that it uses fishamId instead of url to connect to the room.

## Motivation and Context

We update the FishjamRoom component so that it uses the new way of joining the fishjam room.

## How has this been tested?

I've tested this in the fishjam-chat app

## Types of changes

      not work as expected)

## Checklist:


MiloszFilimowski (approved):
LGTM. Great job!

